PR #53, opened 2018-03-02
[density] remove wrong length check

fredRos:
The number of parameters and the number of calls to the density are unrelated. In the unit test, it happened to be 2D and 2 calls. Overly symmetric tests are evil!


oschulz:
Thanks for the catch.

PR #61, opened 2018-11-26
Effective sample size

mszly:
Better definition of the effective sample size that does not incur into negative estimations.


oschulz:
Replaced by #62

PR #67, opened 2019-04-01
Bug fix: plots now use weights from sample.weight

Cornelius-G:


oschulz:
Slightly different version, just commited: d808f7f

PR #70, opened 2019-08-28
1D-Histograms as prior distributions

lmh91:
1-dimensional StatsBase.Histogram's can now be used as prior distributions.

```julia
# h is of type StatsBase.Histogram{<:Real, 1}
d = BAT.HistogramAsUvDistribution(h)
prior = NamedPrior(par1 = d) 
```


oschulz:
Thanks!

PR #71, opened 2019-08-29
Cuba integration is replaced by QuadGK 

VasylHafych:
Relative error tolerance in `QuadGK.quadgk` is changed to 0.001. Initial value used in `Cuba.vegas` was 0.02. There is no loss of performance.


oschulz:
Thanks, @VasylHafych !

PR #73, opened 2019-09-02
N-Dimensional Histograms as Distributions

lmh91:
I added the `struct HistogramAsMvDistribution{T, N}` to create a distribution from a N-dimensional `StatsBase.Histogram{<:Real, N}`. 

I also added some tests.


src/distributions/HistogramAsMvDistribution.jl
```diff
@@ -0,0 +1,139 @@
...
+                c[idim, jdim] += s_inv * (mps[idim][cart_inds[i][idim]] - mean[idim]) * (mps[jdim][cart_inds[i][jdim]] - mean[jdim]) * h.weights[i]
+            end
+        end
+    end
+    return c
+end
+
+
+function Distributions._rand!(::AbstractRNG, d::HistogramAsMvDistribution{T,N}, A::AbstractArray{<:Real,1})::Nothing where {T, N}
+    rand!(A)
```
oschulz: Should use the `AbstractRNG` instance passed as an argument

src/distributions/HistogramAsMvDistribution.jl
```diff
@@ -0,0 +1,139 @@
...
+    cart_inds = CartesianIndices(h.weights)
+    @inbounds for i in eachindex(h.weights)
+        for idim in 1:N
+            m[idim] += s_inv * mps[idim][cart_inds[i][idim]] * h.weights[i]
+        end
+    end
+    return m
+end
+
+function StatsBase.var(h::StatsBase.Histogram{<:Real, N}; T::DataType = Float64, mean = StatsBase.mean(h, T = T), ) where {N}
```
oschulz: This is type piracy, BAT doesn't own `var` or `Histogram`. Just call it `_var` - but we should contribute something like this to StatsBase, eventually.

src/distributions/HistogramAsMvDistribution.jl
```diff
@@ -0,0 +1,139 @@
...
+    cart_inds = CartesianIndices(h.weights)
+    @inbounds for i in eachindex(h.weights)
+        for idim in 1:N
+            v[idim] += s_inv * (mps[idim][cart_inds[i][idim]] - mean[idim])^2 * h.weights[i]
+        end
+    end
+    return v
+end
+
+function StatsBase.cov(h::StatsBase.Histogram{<:Real, N}; T::DataType = Float64, mean = StatsBase.mean(h, T = T)) where {N}
```
oschulz: Also type piracy, just call it `_cov` - see above.

src/distributions/HistogramAsMvDistribution.jl
```diff
@@ -0,0 +1,135 @@
...
+                c[idim, jdim] += s_inv * (mps[idim][cart_inds[i][idim]] - mean[idim]) * (mps[jdim][cart_inds[i][jdim]] - mean[jdim]) * h.weights[i]
+            end
+        end
+    end
+    return c
+end
+
+
+function Distributions._rand!(r::AbstractRNG, d::HistogramAsMvDistribution{T,N}, A::AbstractArray{<:Real,1})::Nothing where {T, N}
+    @inbounds begin
```
oschulz: Let's take this out, such a high-level `@inbounds` is a risky without a lot of checks

oschulz: We can do performance tuning later on. :-)

test/distributions/test_HistogramAsMvDistribution.jl
```diff
@@ -0,0 +1,26 @@
...
+using BAT
+using Test
+
+using Random
+using Distributions, StatsBase
+
+
+@testset "HistogramAsMvDistribution" begin
+    μ = [1.23, -0.67]
+    Σ = transpose([0.45 0.32; 0.32 0.76]) * [0.45 0.32; 0.32 0.76]
```
oschulz: Why not simply `[0.45 0.32; 0.32 0.76]' * [0.45 0.32; 0.32 0.76]` ?

src/distributions/HistogramAsMvDistribution.jl
```diff
@@ -0,0 +1,135 @@
...
+            i = cell_car_index[idim]
+            sub_int = d.edges[idim][i:i+1]
+            sub_int_width::T = sub_int[2] - sub_int[1]
+            A[idim] = sub_int[1] + sub_int_width * A[idim]
+        end
+    end
+    return nothing
+end
+
+function Distributions._rand!(r::AbstractRNG, d::HistogramAsMvDistribution{T,N}, A::AbstractArray{<:Real,2})::Nothing where {T, N}
```
oschulz: Try if

    Distributions._rand!.((r,), (d,), nestedview(A))

works, instead of the for-loop.

oschulz: Use `AbstractMatrix{<:Real})` instead of `AbstractArray{<:Real,2})`, for simplicity.

src/distributions/HistogramAsMvDistribution.jl
```diff
@@ -0,0 +1,135 @@
...
+            for jdim in 1:N
+                c[idim, jdim] += s_inv * (mps[idim][cart_inds[i][idim]] - mean[idim]) * (mps[jdim][cart_inds[i][jdim]] - mean[jdim]) * h.weights[i]
+            end
+        end
+    end
+    return c
+end
+
+
+function Distributions._rand!(r::AbstractRNG, d::HistogramAsMvDistribution{T,N}, A::AbstractArray{<:Real,1})::Nothing where {T, N}
```
oschulz: Use `AbstractVector{<:Real})` instead of `AbstractArray{<:Real,1})`, for simplicity.

src/distributions/HistogramAsMvDistribution.jl
```diff
@@ -0,0 +1,135 @@
...
+    end
+    return v
+end
+
+function _cov(h::StatsBase.Histogram{<:Real, N}; T::DataType = Float64, mean = StatsBase.mean(h, T = T)) where {N}
+    s_inv::T = inv(sum(h.weights))
+    c::Matrix{T} = zeros(T, N, N)
+    mps = StatsBase.midpoints.(h.edges)
+    cart_inds = CartesianIndices(h.weights)
+    @inbounds for i in eachindex(h.weights)
```
oschulz: Shouldn't that be `for i in eachindex(cart_inds)`?

Maybe use `for ... in cart_inds`?

`@inbounds` is too risky at this level without additional checks.

src/distributions/HistogramAsMvDistribution.jl
lmh91: Yep, that works!

oschulz:
Thanks a lot for the contribution!

PR #75, opened 2019-09-17
Array of distributions in Named Priors

lmh91:
```julia
prior = NamedPrior( par1 = [Normal(0.0, 2.0), Normal(1.0, 1.0)] )
```
is now possible.


oschulz:
Nice idea, thanks!

PR #76, opened 2019-09-24
Additional criteria that removes chains with LogLik < -Inf on the generation step is added.

VasylHafych:


oschulz:
Obsoleted by improved implementation.

PR #79, opened 2019-10-23
Use ["cov. weighted result"] uncertainty estimate as a default option for the AHMI integration. 

VasylHafych:


oschulz:
Does this make AHMI run faster, after all?

PR #80, opened 2019-11-19
Prevent AHMI from a crash if trimming is not successful.

VasylHafych:
Condition that prevents AHMI trimming from the deletion of all the integral estimates.


oschulz:
Thanks!

PR #81, opened 2019-12-05
AHMI Bugfixes

VasylHafych:
- AHMI Partial whitening is added. 
- AHMI Tolerance level is restricted to finite values.
- Hard bounds are set to the default value.


oschulz:
Merged manually

PR #112, opened 2020-04-03
Funnel and Multimodal Cauchy distributions for testing bat_integrate

tldrma:


tldrma:
Possible plotting issue using current master branch:

When using BAT to iid sample the multimodal cauchy without truncation, the aspect ratio appears to be incorrect (to the point it seems flat). After truncation, it appears as expect.

VasylHafych:
I had this problem, too. I think this is because Cauchy distribution has an extremely long tail and sometimes you can have a sample that is located hundreds of magnitude orders further from the center of the distribution (if no truncation applied). This destroys the automatic binning of the histogram ...

oschulz:
Ok, let me know when this is ready for review.

src/distributions/funnel.jl
```diff
@@ -0,0 +1,69 @@
...
+
+"""
+The Gaussian Shell distribution
+
+r = 5.0
+w = 2.0
+c = zeros(2)
+"""
+
+function Funnel(a::Real, b::Real, λ::AbstractVector)
```
oschulz, line 18: Use

    function Funnel(;a::Real = ..., b::Real = ..., λ::AbstractVector = 1:5)

to have defaults for all arguments - should be kwargs.

src/distributions/multimodal_cauchy.jl
```diff
@@ -0,0 +1,57 @@
...
+
+"""
+Multimodal Cauchy distribution as described in the AHMI paper.
+Assumes a mixture of two modes.
+
+BAT.MultimodalCauchy(μ=1.0,σ=0.2, n=4)
+
+"""
+
+function MultimodalCauchy(μ::Real, σ::Real, n::Int64)
```
oschulz, line 16: Use kwargs with default values:

    function MultimodalCauchy(μ::Real = ..., σ::Real = ..., n::Integer)

test/integration/test_integration.jl
```diff
@@ -0,0 +1,52 @@
...
+@testset "bat_integrate" begin
+    @testset "multimodal cauchy" begin
+        μ=1.0
+        σ=0.2
+        n_modes=10
+
+        lb = -8
+        ub = 8
+
+        dist = BAT.MultimodalCauchy(μ, σ, n_modes)
```
oschulz, line 16: Use defaults for mu and sigma

test/integration/test_integration.jl
```diff
@@ -0,0 +1,52 @@
...
+                         0. 0. 0. 1. 0.;
+                         0. 0. 0. 0. 1.])
+        sample = bat_sample(dist, 100000).result
+        sample_integral = bat_integrate(sample).result
+
+        @test isapprox(sample_integral, 1., atol=3*sample_integral.err)
+    end
+
+    @testset "funnel" begin
+        dist = BAT.Funnel(1.0, 0.5, [1.0, 2.0, 3.0, 4.0, 5.0])
```
oschulz, line 41: Use default argument values as much as possible

tldrma:
Couple comments before submission for review:

1. Do we want to include basic statistics like mean? I'm not sure if they will be used by anyone.
2. The build is failing due to something called "GKSwstype", which I think is just some graphical stuff. Not sure why mine is failing and the master branch is passing. Safe to ignore?

oschulz:
> Do we want to include basic statistics like mean? I'm not sure if they will be used by anyone.

Yes, please! `mean`, `var` and `cov` should be defined (if they exist for the distribution in question).

oschulz:
> The build is failing due to something called "GKSwstype"

Nope, take a closer look. :-)

```
Warning: 2 docstrings potentially missing:

    BAT.FunnelDistribution :: Tuple{}
    BAT.MultimodalCauchy :: Tuple{}
```

src/distributions/funnel.jl
```diff
@@ -0,0 +1,58 @@
...
+
+*BAT-internal, not part of stable public API.*
+
+Funnel distribution (Caldwell et al.)[https://arxiv.org/abs/1808.08051].
+
+# Arguments
+- `a::Real`: Variance of the dominant normal distribution.
+- `b::Real`: Variance of the supporting normal distributions.
+- `n::Int`: Number of dimensions.
+"""
```
tldrma, line 20: Should this simply be moved above the struct line, and then no docstring needed for the function to construct it?

Same logic applies to MultimodalCauchy?

oschulz: Yes, see some other types in BAT.jl, the ctor function should be documented as part of the docstring of the struct it belongs to.

src/distributions/funnel.jl
```diff
@@ -0,0 +1,64 @@
+# This file is a part of BAT.jl, licensed under the MIT License (MIT).
+
+@doc doc"""
+    BAT.FunnelDistribution([a=1, b=0.5, n=3])
+
+*BAT-internal, not part of stable public API.*
```
oschulz, line 6: Let's make this part of the experimental API, not the unstable API.

src/distributions/multimodal_cauchy.jl
```diff
@@ -0,0 +1,57 @@
+# This file is a part of BAT.jl, licensed under the MIT License (MIT).
+
+@doc doc"""
+    BAT.MultimodalCauchy([μ=1, σ=0.5, n=3])
+
+*BAT-internal, not part of stable public API.*
```
oschulz, line 6: Let's make this part of the experimental API, not the unstable API.

src/distributions/funnel.jl
```diff
@@ -0,0 +1,64 @@
...
+    n = length(λ)
+    a = float(a)
+    b = float(b)
+    dist = Vector{Normal}(undef, n)
+    dist[1] = Normal(0.0, a^2)
+    for i in 2:n
+        @inbounds dist[i] = Normal(0.0, exp(2*b*λ[1]))
+    end
+    return product_distribution(dist)
+end
```
oschulz, line 64: We should add support for `StatsBase.params`, like many dists in Distributions.jl have.

src/distributions/multimodal_cauchy.jl
```diff
@@ -0,0 +1,57 @@
...
+
+function Distributions._rand!(rng::AbstractRNG, d::MultimodalCauchy, x::AbstractVector)
+    Distributions._rand!(rng, d.dist, x)
+end
+
+function _construct_dist(mixture_model::MixtureModel, σ::Real, n::Int64)
+    vector_of_dists = vcat(mixture_model, mixture_model, [Cauchy(0, σ) for i in 3:n])
+    dist = product_distribution(vector_of_dists)
+    return dist
+end
```
oschulz, line 57: We should add support for `StatsBase.params`, like many dists in Distributions.jl have.

test/integration/test_integration.jl
```diff
@@ -0,0 +1,38 @@
+using Distributions
+using BAT
+using ValueShapes
+using IntervalSets
+using Test
+
+@testset "bat_integrate" begin
+    @testset "multimodal cauchy" begin
+        dist = BAT.MultimodalCauchy()
```
oschulz, line 9: We can factor this out into a function `test_integration(title, dist, nsamples, val_expected, val_rtol, err_max)` or so, and then use `test_integration("multimodal cauchy", BAT.MultimodalCauchy(), 10^5, 1, ..., 0, 0.15)`.

`test_integration` should be defined inside of `@testset "bat_integrate"`.

test/integration/test_integration.jl
```diff
@@ -0,0 +1,38 @@
...
+        dist = BAT.MultimodalCauchy() 
+
+        sample = bat_sample(dist, 100000).result
+        sample_integral = bat_integrate(sample).result
+
+        @test isapprox(sample_integral.val, 1, atol=3.1*sample_integral.err)
+        @test sample_integral.err < 0.15
+    end
+
+    @testset "MvN" begin
```
oschulz, line 18: Can use a function like `test_integration` (see above) to avoid code duplication.

test/integration/test_integration.jl
```diff
@@ -0,0 +1,38 @@
...
+                         0. 0. 1. 0. 0.;
+                         0. 0. 0. 1. 0.;
+                         0. 0. 0. 0. 1.])
+        sample = bat_sample(dist, 100000).result
+        sample_integral = bat_integrate(sample).result
+
+        @test isapprox(sample_integral, 1., atol=3.1*sample_integral.err)
+    end
+
+    @testset "funnel" begin
```
oschulz, line 30: Can use a function like `test_integration` (see above) to avoid code duplication.

test/integration/test_integration.jl
oschulz: Probably should add an `algorithm` argument to `test_integration` as well (maybe as first argument), and specify `AHMI()` explicitly. We'll want to test other integrators too, in the future, but not all will be able to handle all test functions.

tldrma: To the first point then, we would want something like
`
@tesetset "bat_integrate" begin
    function test_integration(title::String, dist::Distribution, nsamples::Int, val_expected::Real=1.0, val_rtol::Real=3.1, err_max::Real=0.1)
        @testset title begin
            sample = bat_sample(dist, nsamples).result
            sample_integral = bat_integrate(sample).result
            
        @test isapprox(sample_integral.val, val_expected, val_rtol*sample_integral.err)
        @test sample_integral.err < err_max
    end 
`

oschulz: Yes, try that, should make the code more compact.

oschulz:
Looks good, but you have test failures on 32-bit. Replace all `n::Int64` arguments by `n::Integer`.

In general, `Int` should be used instead of `Int64`, instead in cases where 64-bits are really needed. And for function arguments, the generic type `Integer` is almost always preferable.

oschulz:
No, really, use `Integer` for function arguments, not `Int` (except in special cases). :-) For type fields it would be the other way round, `Int` instead of `Integer`.

oschulz:
You still need to fix the 32-bit issues! :-)

tldrma:
> You still need to fix the 32-bit issues! :-)

I thought you were saying to make the types inside of a struct Int, and the ctor args Integer, but I think you were actually saying they should both be Integer? I'm noticing that DataSet uses Integer, and not Int.

oschulz:
Ah, sorry - it's not in your code anymore, it's in the DataSet ctor. Change the `nsubsets::Int64 = 0` to `nsubsets::Integer = 0` in the DataSet ctors.

tldrma:
I changed the two DataSet ctor args from Int to Integer (commit 8dcc09dcd3890db25120c1df30b35f7672f22275), but the error persisted.
My understanding is that Int will default to the correct bit size, so I tried changing the Int64 types to Int (commit fe4c712ac09e98e0b842f59d687c601e0d19289a), and it leads to similar problems in a different part of the code. Am I incorrect in what I'm doing?

oschulz:
No, you're doing the right thing - it's just that a lot of internal AHMI functions have explicit `Int64` arguments - but they shouldn't. In this case `reduced_volume_hm` ... you'll have to follow the breadcrumbs and clean up that kind of stuff on the way. :-)

tldrma:
That's fine with me. I just wanted to check before going down this path. Thanks!

src/distributions/funnel.jl
```diff
@@ -0,0 +1,60 @@
...
```
tldrma, line 60: Just as an aside, I tried to remove the _construct_dist function here, and instead moved the code into _logpdf() directly. Here's a benchmark for comparison. I'm not sure how big of a deal it is that there's this highly specific function just floating around in funnel.jl, so just let me know if removing it is preferred.

Using 10^5 samples...
```
# With _construct_dist()

julia> @benchmark pdf(funnel, data) samples=100 seconds=50
BenchmarkTools.Trial: 
  memory estimate:  116.72 MiB
  allocs estimate:  2899491
  --------------
  minimum time:     100.158 ms (7.12% GC)
  median time:      107.260 ms (10.18% GC)
  mean time:        108.262 ms (10.14% GC)
  maximum time:     160.009 ms (14.33% GC)
  --------------
  samples:          100
  evals/sample:     1


# Without _construct_dist()

julia> @benchmark pdf(funnel, data) samples=100 seconds=50
BenchmarkTools.Trial: 
  memory estimate:  223.53 MiB
  allocs estimate:  5999491
  --------------
  minimum time:     193.308 ms (5.68% GC)
  median time:      198.714 ms (8.26% GC)
  mean time:        204.271 ms (8.85% GC)
  maximum time:     321.480 ms (38.47% GC)
  --------------
  samples:          100
  evals/sample:     1
```

Can also check how it scales if needed

oschulz: Having a helper function is fine, can make debugging easier.

Could be that you have some type instability in there - in such cases, an additional dispatch can help with performance. See if you can find (e.g. with `@code_warntype`) and eliminate it. :-)

oschulz: One millisecond per sample seems really slow, will be hard to run tests this way.

oschulz:
Just resolved some conflicts, if the tests pass, I'll merge right away.

src/distributions/funnel.jl
tldrma: For some reason `@code_warntype` only infers `Any` for the output. I tried to annotate the entire function, but it didn't seem to do much. I'll look into it more.

oschulz:
Done. :-)

src/distributions/funnel.jl
oschulz: It may be due to some other function called from there, you may have to follow breadcrumbs again ... :-)

PR #123, opened 2020-05-22
Fix bug in plotting of 1D samples

Cornelius-G:


oschulz:
@Cornelius-G, could you check why the docs-generation fails?

Cornelius-G:
@oschulz I have no idea why the docs fail again. It were only minor changes that should not affect the docu. However, the error message is `ERROR: LoadError: ArgumentError: Package Documenter [e30172f5-a6a5-5a46-863b-614d45cd2de4] is required but does not seem to be installed:` So it seems there is a problem when Travis is setting up the environment. Is it possible to just retry building the docs?

Cornelius-G:
Ok, it seems it automatically restarted and now passed.

oschulz:
I ran it again - looks like it was just a temporary failure. Sorry, should have checked first.

PR #124, opened 2020-05-22
Add HMC sampler

Cornelius-G:
First version of HMC sampler using AdvancedHMC


oschulz:
Thanks a lot for this!

PR #128, opened 2020-05-28
Function to estimate modes in marginalizations

VasylHafych:
The function returns local mode given posterior samples. The optimal number of bins can be determined using [this](https://github.com/JuliaPlots/Plots.jl/blob/8859518f2d60eb97118a7e245c7b6f8fed8e4bc7/src/recipes.jl#L665) implementation from Plots.jl.


src/plotting/recipes_samples_overview.jl
```diff
@@ -108,3 +108,110 @@
     end
 
 end
+
+@recipe function f(model::Function,
```
VasylHafych, line 112: Hi @Cornelius-G! I have added a plotting recipe that allows us to show error bars on the fit parameters.  I have never worked with plotting recipes though, so can you please review the code and give your feedback?

VasylHafych:
Hi @oschulz! I have extended one of the plots from the tutorial with a new plotting recipe. It is very easy to evaluate it and it makes the plot ["Data, True Model and Best Fit"](https://github.com/VasylHafych/BAT.jl/blob/0efe85ccac9e58fa8090fbf50ebf9b1d2de75ef9/docs/src/tutorial_lit.jl#L379) better.

oschulz:
We should change the plotting code to use the new `bat_findmode`, so that we don't duplicate that code.

Cornelius-G:
@oschulz But it seems the new function just calls the old `find_localmodes` from plotting. So no duplication

src/plotting/recipes_samples_overview.jl
Cornelius-G: Hi @VasylHafych, thanks for the very nice recipe. I will take a closer look into the code.

VasylHafych:
Indeed, there is no code duplication. I call old `find_localmodes` function from plotting recipes.

oschulz:
Oh, yes - then can should move that code into the new exported function and remove the old one, right?

oschulz:
Within the scope of this PR, let's replace the `BATHistogram` functions by a new, exported and generic `bat_marginalize()` function that follows the scheme of the other `bat_...` functions (return a `(result=..., )`).

We can rename the type `BATHistogram` to `Marginalization` - that type should not be exported, for now. Next to the histogram that `BATHistogram` currently contains, it should also contain a reference to the parameter(s), so that the plotting recipe for it doesn't need the separate `idx` argument anymore.

oschulz:
Which reminds me: We shouldn't use the term "local mode" anywhere in this context any longer (also in the plots). A local mode is a "bump" in a multi-modal distribution, lower than the global mode. But here's we're talking about global modes in the marginalizations of the posterior.

src/plotting/recipes_samples_overview.jl
```diff
@@ -108,3 +108,110 @@
...
+        x::Union{StepRangeLen, Vector},
+        samples::DensitySampleVector;
+        conf_intervals = standard_confidence_vals,
+        colors = standard_colors,
+        mean = true,
+        globalmode = true,
+        localmode = true,
+        xlabel = "x",
+        ylabel = "f(x)",
+        title="",
```
Cornelius-G, line 122: `xlabel`, `ylabel`, `title`, `legend` and `size` are automatically available to the user through the plot recipes. no need to specify them here as keyword arguments.
The default values can be set directly in 143-147: e.g. `xguide --> "x"`  (the `-->` operator automatically  uses the default or if provided the user input )

Cornelius-G:
> Which reminds me: We shouldn't use the term "local mode" anywhere in this context any longer (also in the plots). A local mode is a "bump" in a multi-modal distribution, lower than the global mode. But here we're talking about global modes in the marginalizations of the posterior.

Ok, so maybe call it "marginalized mode(s)" ?

src/plotting/recipes_samples_overview.jl
```diff
@@ -108,3 +108,110 @@
...
+
+    xguide --> xlabel
+    yguide --> ylabel
+    title --> title
+    legend --> legend
+    size --> size
+
+    @series begin
+        ribbon --> (y_ribbons[:,5],y_ribbons[:,6])
+        fillcolor --> colors[1]
```
Cornelius-G, line 151: It seems the colors are inverted compared to our usual convention. So the innermost ribbon should be green, the outermost red.

src/plotting/recipes_samples_overview.jl
```diff
@@ -108,3 +108,110 @@
...
+    y_ribbons = zeros(length(x), 6)
+    y_50_interval = zeros(length(x))
+
+    intervals = 0.5*(1 .- conf_intervals)
+
+    for x_ind in Base.OneTo(length(x))
+        y_samples = model.(samples.v, x[x_ind])
+
+        y_50_interval[x_ind] = quantile(y_samples, 0.5)
+        y_ribbons[x_ind,:] .= [quantile(y_samples, intervals[1]), quantile(y_samples, 1. - intervals[1]),
```
Cornelius-G, line 136: This should actually be automatized to use as many ribbons as the user specifies in the keyword argument `conf_intervals`.

src/plotting/recipes_samples_overview.jl
```diff
@@ -108,3 +108,110 @@
...
+            linestyle --> :dot
+            linewidth --> 1.5
+            label --> "Global Mode"
+            x, broadcast(x -> model(global_mode_params, x), x)
+        end
+    end
+
+    if mean
+
+        global_mode_params = mode(samples)[1]
```
Cornelius-G, line 206: is this used anywhere below?

src/plotting/recipes_samples_overview.jl
```diff
@@ -108,3 +108,110 @@
...
+        @series begin
+            linecolor --> :black
+            linestyle --> :dot
+            linewidth --> 1.5
+            label --> "Global Mode"
+            x, broadcast(x -> model(global_mode_params, x), x)
+        end
+    end
+
+    if mean
```
Cornelius-G, line 204: Since it is the 0.5 quantile, this should rather be called "median"

oschulz:
Yes, or "marginal modes" or so - I sent an email.

src/plotting/recipes_samples_overview.jl
```diff
@@ -108,3 +108,110 @@
...
+        ribbon --> (y_ribbons[:,5],y_ribbons[:,6])
+        fillcolor --> colors[1]
+        linecolor --> colors[1]
+        seriesalpha --> 1
+        linealpha --> 0.5
+        label --> "$(standard_confidence_vals[3])"
+        x, y_50_interval
+    end
+
+    @series begin
```
Cornelius-G, line 159: this should also be automatized, referring to line 136. You can just make a `for` loop containing `@series begin ... end`.

Cornelius-G:
@VasylHafych I just used the "review" feature to give some comments on the plot recipe. I hope you can work with that. I used this feature for the first time, so just ask me if there is something unclear. The comments are just some possible improvements. But generally this is quite nice work!

Cornelius-G:
> Within the scope of this PR, let's replace the `BATHistogram` functions by a new, exported and generic `bat_marginalize()` function that follows the scheme of the other `bat_...` functions (return a `(result=..., )`).
> 
> We can rename the type `BATHistogram` to `Marginalization` - that type should not be exported, for now. Next to the histogram that `BATHistogram` currently contains, it should also contain a reference to the parameter(s), so that the plotting recipe for it doesn't need the separate `idx` argument anymore.

Ok, this seems generally like a good idea, but I have some questions on how to realize this.
1. What should `bat_marginalize()` return? A `Marginalization`? As a user I would rather expect this to give me the samples of the specified dimensions. 
2. The idea of `BATHistogram` was also to provide the BAT plotting recipes for other histograms, that have not been created from samples. So basically every `StatsBase.Histogram` could be converted to `BATHistogram` to provide plots like `smallest_intervals` etc. to make it easier for users to plot their data in the same style. When we rename this to `Marginalization` this does not make so much sense.

oschulz:
> What should bat_marginalize() return? A Marginalization? As a user I would rather expect this to give me the samples of the specified dimensions.

Good point. I believe it should return a marginal distribution. Let's replace BATHistogram by a new type MarginalDist, which would contain a `Distribution` (single- or multivariate) and a reference to the parameter(s) that remain after marginalization. The Distribution would then usually be a `UvBinnedDist` or `MvBinnedDist` from ["EmpiricalDistributions.jl"](https://github.com/oschulz/EmpiricalDistributions.jl), which wrap a histogram in term.

This way, things would be nice an clean: The result of a marginalization could immediate be reused as a prior, and at the same time all necessary information would be available to plotting recipes, without type piracy.

In the future, we can then add non-histogrammed, sample-based Distributions to "EmpiricalDistributions.jl", and offer different marginalization algorithms to allow the user to select which kind of marginalization they want.

The `bat_marginalmode` function (or whatever we'll call it) can then use this machinery internally, and take the marginialization algorithm as an optional argument (like the other `bat_...` functions). The rest will be automatic, since the empirical distribution will already implement a `mode` function (we have to add this to `UvBinnedDist` and `MvBinnedDist`).

oschulz:
Let's go with the term "marginal mode".

src/plotting/recipes_samples_overview.jl
VasylHafych: Thanks, corrected!

src/plotting/recipes_samples_overview.jl
VasylHafych: Corrected!

src/plotting/recipes_samples_overview.jl
VasylHafych: Fixed! By the way, I was always quite confused by this choice of colors. For me, it would be much more natural to use intense and bright red in a center (demonstrating a high density) and light green on the periphery (showing low density). For example, look at [this](https://seaborn.pydata.org/examples/joint_kde.html) Python Seaborn package.

VasylHafych:
Hi @Cornelius-G. Thank you for your helpful suggestions. I have included an update function in my latest commit.

@oschulz, considering that your suggestions with renamings require more work to be done, I propose to move my plotting recipe to a separate pull request to decouple these two topics.

oschulz:
> I propose to move my plotting recipe to a separate pull request to decouple these two topics.

Sure!

VasylHafych:
@oschulz, a new pull request is ready. 

@Cornelius-G, if you could rename `BATHistogram` from your Plotting recipes by a new type `MarginalDist` as Oliver [suggested](https://github.com/bat/BAT.jl/pull/128#issuecomment-635904616), I can then use your updates and add two functions: `bat_marginalmode`, `bat_median`.

One more question — should we rename these functions in BAT 2.0 to have name consistency? They all are acting on posterior and returning a DensitySampleVector. For example, we can call them:
* `bat_globalmode`
* `bat_marginalmode`
* `bat_median`
* `bat_mean`.

oschulz:
> @oschulz, a new pull request is ready

Thanks! I just posted some comments on #129 .

oschulz:
> if you could rename BATHistogram from your Plotting recipes by a new type MarginalDist

Yes - it's not just a rename, though.

Cornelius-G:
>Let's replace BATHistogram by a new type MarginalDist, which would contain a `Distribution` (single- or multivariate) and a reference to the parameter(s) that remain after marginalization. The Distribution would then usually be a `UvBinnedDist` or `MvBinnedDist` from ["EmpiricalDistributions.jl"](https://github.com/oschulz/EmpiricalDistributions.jl), which wrap a histogram in term.

Ok, I guess this will be on my agenda then. 
@oschulz could you maybe give me an example of how exactly you want this `MarginalDist` to look like. Especially the parameter reference? I would then try to change the plotting accordingly.

oschulz:
I guess it would be something like

```julia
struct MarginalDist{N,D<:Distribution, VS<:AbstractValueShape}
    dist::D,
    origvalshape::VS
    dims::NTuple{N,Int}
end
```

with `dims` referencing the dimensions (in unshaped par-space) left after marginalization.

VasylHafych:
> Thanks! I just posted some comments on #129 .

Cannot see any of your comments on #129

oschulz:
> Cannot see any of your comments on #129

Search the page for "oschulz started a review"

VasylHafych:
> oschulz started a review

0 results are matched. I guess it is not visible for me or something.

oschulz:
Oh, sorry, I forgot to click "submit review" :-)

VasylHafych:
I was playing around with [KernelDensity.jl](https://github.com/JuliaStats/KernelDensity.jl) and realized that we can use it to construct KDE of our marginal distributions and then use an optimizer to find marginal modes. In this case, we should be able to find marginal mode with better precision than just doing binning. What do you think about that?

oschulz:
Yes, I've been thinking about KDE, too - that should go into the EmpiricalDistributions package though. We could add KDE-based empirical distributions, in addition to the bin-based ones. That would be very valuable! Especially because they would be differentiable, so suitable to use as priors with HMC and so on - bin-based step-function priors wouldn't do so well there.

VasylHafych:
@oschulz, Yes, it sounds like an interesting task. I have already played a bit with Julia KDE, so I can volunteer to implement `EmpiricalDistributions` extension. But I cannot say exactly when -  maybe a little bit later.

oschulz:
> But I cannot say exactly when - maybe a little bit later.

Sure, there's no rush - the nice thing is, if we do things properly in BAT, something like that can be added to EmpiricalDistributions later on and will then "just work".

oschulz:
I just merged the other PR by @Cornelius-G - could you adept this PR to the changes, @VasylHafych ?

VasylHafych:
Hi @oschulz. Here is an updated version. We now have `bat_findmode`, `bat_findlocalmode`, and `bat_findmedian`.  The last two functions operate on `DensitySampleVector`.

oschulz:
> We now have bat_findmode, bat_findlocalmode, and bat_findmedian

I think we decided to call this marginal mode, not local mode, right? Or is `bat_findlocalmode` really about actual local modes in the non-marginalized samples?

VasylHafych:
> I think we decided to call this marginal mode, not local mode, right? 

sorry, this name remained from the old implementation. The function is now called `bat_marginalmode`. Also, I added `Statistics.median` (see last commit). 

Ready to merge.

oschulz:
Sorry, we a little change in the output of `bat_marginalize` that you'll have to adapt to, first: #133

oschulz:
Thanks, will merge as soon as tests are through.

oschulz:
Thanks again!

PR #130, opened 2020-06-08
Introduce MarginalDist

Cornelius-G:
Introduce MarginalDist and remove BATHistogram as discussed in #128


oschulz:
Nice work! Is this ready to merge (if tests pass)?

Cornelius-G:
I think so. I'm just not sure if there might be conflicts with @VasylHafych's new changes.
Unfortunately, it is now no longer so straightforward to plot (StatsBase) histograms with the BAT plotting styles, which is for example helpful when comparing analytical results to BAT sampling results. But maybe we can think of a nice solution for this. Maybe even by including some of the recipes in EmpiricalDistributions.jl?

oschulz:
> Unfortunately, it is now no longer so straightforward to plot (StatsBase) histograms with the BAT plotting styles

How about adding a `MarginalDist(hist::Histogram)` constructor, to make this more user friendly?

We should also decide what `MarginalDist` will be, a `Distribution` or a `Density`. We we can do that in another PR.

> Maybe even by including some of the recipes in EmpiricalDistributions.jl?

Sure!

oschulz:
I've been thinking (doesn't have to happen in this PR) - should we switch from `a[i]` to `a_i` to index into vector-valued parameters? We wouldn't need expression, then - also, we could build proper new varshapes for `MarginalDist`, that only include the parameters that are left, with valid parameter names. Should make it easier to reuser marginalizations as priors.

Cornelius-G:
> I've been thinking (doesn't have to happen in this PR) - should we switch from `a[i]` to `a_i` to index into vector-valued parameters? We wouldn't need expression, then - also, we could build proper new varshapes for `MarginalDist`, that only include the parameters that are left, with valid parameter names. Should make it easier to reuser marginalizations as priors.

Well, we did use this underscore solution in the first implementation a couple weeks ago, so we could go back to this style and use just symbols again. But both solutions have pros and cons. It is indeed easier to enter symbols with the underscore than using the expressions. On the other hand, users would have to be careful to not implement two priors with similar names, like `a_1` and `a_2`. Or were there any further reasons than this why we changed from underscore to expressions?

oschulz:
I think we changed this before we thought about a new, clean way to do marginalizations ... :-)

Hm, I guess the only trouble would be if a user had both `a` and `a_1` in the parameter names ... maybe we should just throw an error in that case. We should still allow the user to use `a[1]`, to specify what to plot/marginalize - the question is, should we also allow `a_1`, and which of the two possibilities will we use in the generated plots/marginalization?

VasylHafych:
> I think so. I'm just not sure if there might be conflicts with @VasylHafych's new changes.
> Unfortunately, it is now no longer so straightforward to plot (StatsBase) histograms with the BAT plotting styles, which is for example helpful when comparing analytical results to BAT sampling results. But maybe we can think of a nice solution for this. Maybe even by including some of the recipes in EmpiricalDistributions.jl?

No, it should not conflict with my changes.

oschulz:
@VasylHafych , PR #128 has conflicts in

* docs/src/tutorial_lit.jl
* src/plotting/recipes_samples_overview.jl

VasylHafych:
Resolved.

PR #131, opened 2020-06-09
Add first versions of sobol, grid and prior samplers

Cornelius-G:


src/samplers/importance/importance_sampler.jl
```diff
@@ -0,0 +1,49 @@
...
+struct SobolSampler <: ImportanceSampler end
+struct GridSampler <: ImportanceSampler end
+struct PriorSampler <: ImportanceSampler end
+
+
+function bat_sample(
+    posterior::AnyPosterior,
+    n::AnyNSamples,
+    algorithm::ImportanceSampler;
+    bounds::Vector{<:Tuple{Real, Real}} = get_prior_bounds(posterior)
```
oschulz, line 14: Let's use `bounds::Any` here, and use `convert(BAT.HyperRectBounds, bounds)` inside. We may have to add a few `convert` methods to `BAT.HyperRectBounds`, the ctors are already there.

src/samplers/importance/prior_bounds.jl
```diff
@@ -0,0 +1,24 @@
+function get_bounds(prior::NamedTupleDist)
```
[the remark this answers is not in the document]
oschulz: We probably don't need a separate "prior_bounds.jl", this should probably be added to "parambounds.jl".

src/utils/sample_utils.jl
```diff
@@ -1,14 +1,16 @@
 export convert_to_bat_samples
 
-function convert_to_bat_samples(samples, posterior)
-    samples, weights = get_weighted_samples(samples)
-    n = length(samples)
+function convert_to_bat_samples(samples::Any, posterior::Any)
```
oschulz, line 3: What does `convert_to_bat_samples` do? If it's exported, we need to document it. But defining another ctor `DensitySampleVector(...)` is probably better. For that, `(samples::Any, posterior::Any)` is too generic, though.

src/samplers/importance/prior_bounds.jl
```diff
@@ -0,0 +1,24 @@
...
+    bounds = [get_bounds(d) for d in values(prior)]
+    return vcat(bounds...)
+end
+
+
+function get_bounds(d::Distribution)
+    lo, hi = minimum(d), maximum(d)
+
+    lo == -Inf ? lo = quantile(d, 0.00001) : nothing
```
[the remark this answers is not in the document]
Cornelius-G: right, what is the advantage of your implementation?

src/samplers/importance/prior_bounds.jl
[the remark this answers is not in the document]
Cornelius-G: Ok, I will try to put this into the `BAT.var_bounds` format. I saw that there is this `BAT.var_bounds` but wasn't sure if this is really what we wanted, so I added this `prior_bounds.jl` as an interim solution.

src/utils/sample_utils.jl
Cornelius-G: @oschulz  `convert_to_bat_samples` was originally there to convert plain samples, so just a vector of floats or so (e.g. as obtained when running standalone AdvancedHMC), into a shaped `DensitySampleVector`. I now changed this to accepted sampling points and evaluate the posterior there, assigning the posterior value as sample weight for the importance samplers. I guess we should find a better name for this. Do you think we should export this? And maybe also add again the function that takes plain "float" samples to convert them to BAT samples? Might be nice to offer this to users, so they can use other samplers but BATs infrastructure for plotting, integration and so on...

src/samplers/importance/prior_bounds.jl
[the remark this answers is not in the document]
oschulz: Type stability.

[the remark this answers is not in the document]
Cornelius-G: Ok, but `isa(Inf, Float64)` is true. So `quantile` is not always a float?

src/utils/sample_utils.jl
oschulz: > And maybe also add again the function that takes plain "float" samples to convert them to BAT samples

Yes, why don't we add ctors

```
DensitySampleVector(
    v::AbstractVector,
    logd::AbstractVector{<:Real},
    weight::AbstractVector{<:Real} = fill(1, length(eachindex(v))),
    info::AbstractVector = fill(nothing, length(eachindex(v))),
    aux::AbstractVector = fill(nothing, length(eachindex(v)))
)

DensitySampleVector(
    v::AbstractVector,
    logd::AbstractDensity,
    weight::AbstractVector{<:Real} = fill(1, length(eachindex(v))),
    info::AbstractVector = fill(nothing, length(eachindex(v))),
    aux::AbstractVector = fill(nothing, length(eachindex(v)))
)
```

That should cover both use cases, right?

src/samplers/importance/prior_bounds.jl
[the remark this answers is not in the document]
oschulz: We shouldn't assume that different functions like `minimum` and `quantile` return the same float type. In practice it's likely to be `Float64` in most cases, but it doesn't have to be.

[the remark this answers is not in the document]
Cornelius-G: Ah, ok. I see

src/utils/sample_utils.jl
Cornelius-G: Ah, yes. This seems like the cleanest solution. I will try to do it like this.

Cornelius-G: Hm.. I'm actually not sure how to make this nice for a general use-case and for our importance sampling.

For the importance samplers, we would need something like
```
function DensitySampleVector(
    v::AbstractVector,
    posterior::AnyPosterior;
    weight::AbstractVector{<:Real} = fill(1, length(eachindex(v))),
    info::AbstractVector = fill(nothing, length(eachindex(v))),
    aux::AbstractVector = fill(nothing, length(eachindex(v)))
)
    logval = density_logval.(Ref(posterior), v)
    weight = exp.(logval)

    shape = varshape(posterior)

    return shape.(DensitySampleVector((ArrayOfSimilarArrays(v), logval, weight, info, aux)))
end
```
But here the weights are some kind of a special case. We could just pass the right weights to the keyword in the `DensitySampleVector` constructor, but then we would again need a function around this.

And for the general use case: When `logd` is just a `AbstractDensity` (as proposed above) we can't get a shape. So should we maybe rather accept a `ValueShape` together with an array of samples? 
Also, how do we handle the option to use (repetition) weights? There is a very preliminary function for removing duplicate samples and counting them for weights (remnant from first HMC implementation). But we can't just pass this to the `weight` keyword, since duplicate samples need to be removed from the samples array.

oschulz: I think the importance sampler should do this internally, and then call the constructor with the finished samples, weights and logval. We shouldn't put an algorithm inside of a constructor. The ctor's will still be useful on their own, though.

Cornelius-G: Ok, so I added a new constructor for `DensitySampleVector`:

```
function DensitySampleVector(
    v::AbstractVector,
    shape::AbstractValueShape;
    logval::AbstractVector{<:Real} = fill(0, length(eachindex(v))),
    weight::Union{AbstractVector{<:Real}, Symbol} = fill(1, length(eachindex(v))),
    info::AbstractVector = fill(nothing, length(eachindex(v))),
    aux::AbstractVector = fill(nothing, length(eachindex(v)))
)
    if weight == :RepetitionWeight
        v, weight = repetition_weights(v)
    end

    return shape.(DensitySampleVector((ArrayOfSimilarArrays(v), logval, weight, info, aux)))
end
```

src/samplers/importance/importance_sampler.jl
Cornelius-G: I modified this to accept `BAT.HyperRectBounds` or `Any`. For user inputs of types `Vector{<:Tuple}` or `Tuple{Vararg{Tuple}}` the bounds will be converted to `BAT.HyperRectBounds`.

src/samplers/importance/prior_bounds.jl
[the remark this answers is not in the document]
Cornelius-G: @oschulz I moved the functions from `prior_bounds.jl` to `distribution_density.jl`, where a lot of the `param_bounds` functions are located. Since I think we need something slightly different than the current `param_bounds` implementation  provides, I named the functions `finite_param_bounds` for now. But I guess we can make these bounds a bit more coherent in the future. I was just not quite sure how to do this right now.

src/utils/sample_utils.jl
oschulz: `shape` should not be a separate argument - users should apply the shape separately, if necessary, e.g. via `DensitySampleVector(shape.(v), ...)`

src/samplers/importance/importance_sampler.jl
oschulz: Hm, I think we should introduce a `TruncatedDensity`, that would probably be the cleanest solution. We'll need that anyhow for space-partitioned sampling soon. So, instead of a separate `bounds` argument, the user would use `TruncatedDensity(posterior, bounds)` as the target distribution.

src/samplers/importance/prior_bounds.jl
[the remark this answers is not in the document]
oschulz: It the right point in time to overhaul the whole param bounds still anyhow. But I'd like to have two different of bound APIs, so why don't we fix what's wrong with the current scheme right now. :-)

What do you need that  `HyperRectBounds` doesn't provide? Would it be enough to just define `Base.isinf` for it or so?

src/densities/distribution_density.jl
```diff
@@ -65,3 +65,45 @@ function dist_param_bounds(d::EmpiricalDistributions.MvBinnedDist{T, N}, bounds_
...
+
+function finite_param_bounds(ntd::NamedTupleDist, bounds_type::BoundsType)
+    bounds = vcat([finite_param_bounds(d) for d in values(ntd)]...)
+    lo = [b[1] for b in bounds]
+    hi = [b[2] for b in bounds]
+    HyperRectBounds(lo, hi, bounds_type)
+end
+
+
+function finite_param_bounds(d::Distribution)
```
oschulz, line 78: This should be restricted to univariate distributions

src/samplers/importance/prior_bounds.jl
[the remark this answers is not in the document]
Cornelius-G: We would just need something that returns `HyperRectBounds` that have no `Inf` values for the bounds. So something like a switch `finite=true` or so in `param_bounds` that then provides bounds with all finite values, even for unlimited distributions. In the simplest case that could be done using something like the quantile estimate we implemented, but as discussed in the meeting, this can of course be improved to provide better estimates for the bounds.

src/densities/distribution_density.jl
```diff
@@ -65,3 +65,45 @@ function dist_param_bounds(d::EmpiricalDistributions.MvBinnedDist{T, N}, bounds_
...
+end
+
+
+function finite_param_bounds(d::Product)
+    bounds = finite_param_bounds.(d.v)
+    return vcat(bounds...)
+end
+
+
+function convert2HyperRectBounds(bounds::Union{Vector{<:Tuple}, Tuple{Vararg{Tuple}}})
```
oschulz, line 99: Let's use `AbstractInterval`s and vectors of `AbstractInterval` instead of 2-tuples for bounds, it's clearer.

Instead of `convert2HyperRectBounds` we can define `HyperRectBounds(bounds)` and `Base.convert(::Type{HyperRectBounds}, bounds).

src/parameters/density_sample.jl
```diff
@@ -157,6 +157,22 @@ DensitySampleVector(::Type{S}, varlen::Integer) where {P<:AbstractVector{<:Real}
     DensitySampleVector{P,T,W,R,Q}(undef, 0, varlen)
 
 
+function DensitySampleVector(
+    v::AbstractVector,
+    shape::AbstractValueShape;
```
oschulz, line 162: Let's not have `shape` in here, better to use `DensitySampleVector(shape.(v), ...)`

src/utils/sample_utils.jl
```diff
@@ -1,32 +1,18 @@
...
-    info = Vector{Nothing}(undef, n)
-    aux = Vector{Nothing}(undef, n)
-
-    varshape(posterior).(DensitySampleVector((ArrayOfSimilarArrays(samples), logval, weights, info, aux)))
-end
-
-
-function get_weighted_samples(samples)
-    n = length(samples)
+function repetition_weights(v::AbstractVector)
```
oschulz, line 1: Maybe call this `repetition_to_weights`?

oschulz: This can go into "density_sample.jl"

src/samplers/importance/prior_bounds.jl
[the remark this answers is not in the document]
oschulz: How about something like this:

```
orig_bounds = var_bounds(posterior)
if isinf(orig_bounds)
    TruncatedDensity(posterior, estimate_finite_bounds(posterior))
else
   TruncatedDensity(posterior, orig_bounds)
end
```

Would be type-stable, too.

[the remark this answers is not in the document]
Cornelius-G: Ah, yes this seems like a nice and clean solution. But we don't have this `TruncatedDensity` yet, right?

src/samplers/importance/importance_sampler.jl
oschulz: No, but you can add that it to this PR - it can be very simple, initially, just something like

```julia
struct TruncatedDensity{D<:AbstractDensity,B<:VarVolumeBounds} <: AbstractDensity
    density::D
    bounds::B
end

Base.parent(density::TruncatedDensity) = density.density

var_bounds(density::TruncatedDensity) = density.bounds

density_logval(density::TruncatedDensity, v::AbstractVector{<:Real}) =
    density_logval(parent(density), v)
```

Should go into it's own file "densities/truncated_density.jl", We'll probably want to add more functionality later on.

oschulz: And you'll need to define `Base.isinf` for bounds in general.

src/densities/posterior_density.jl
```diff
@@ -162,6 +162,11 @@ function var_bounds(density::AbstractPosteriorDensity)
 end
 
 
+function estimate_finite_bounds(posterior::AbstractPosteriorDensity; bounds_type::BoundsType=hard_bounds)
```
oschulz, line 165: We should add a

```julia
"""
    estimate_finite_bounds(...)

*BAT-internal, not part of stable public API.*
...
"""
function estimate_finite_bounds end
```

to "abstract_density.jl", to have the new function (and what it does) defined in a central place. Also needs to be added to "internal_api.md".

src/densities/truncated_density.jl
```diff
@@ -0,0 +1,13 @@
+# This file is a part of BAT.jl, licensed under the MIT License (MIT).
+
+struct TruncatedDensity{D<:AbstractDensity,B<:VarVolumeBounds} <: AbstractDensity
```
oschulz, line 3: This deserves a docstring, with "*BAT-internal, not part of stable public API.*" in it, for now. I think at some point this will become part of the stable public API.

src/parameters/density_sample.jl
```diff
@@ -157,6 +157,21 @@ DensitySampleVector(::Type{S}, varlen::Integer) where {P<:AbstractVector{<:Real}
     DensitySampleVector{P,T,W,R,Q}(undef, 0, varlen)
 
 
+function DensitySampleVector(
+    v::AbstractVector;
+    logval::AbstractVector{<:Real} = fill(0, length(eachindex(v))),
```
oschulz, line 162: Shouldn't logval be required? If we make it optional, we should fill it with `NaN`, but are there really cases where we want to allow the existence  of a `DensitySampleVector` without valid `logval`?

src/parameters/density_sample.jl
```diff
@@ -157,6 +157,21 @@ DensitySampleVector(::Type{S}, varlen::Integer) where {P<:AbstractVector{<:Real}
     DensitySampleVector{P,T,W,R,Q}(undef, 0, varlen)
 
 
+function DensitySampleVector(
```
oschulz, line 160: We should add this new ctor to the docstring of DensitySampleVector

src/parameters/density_sample.jl
```diff
@@ -311,3 +326,31 @@ function drop_low_weight_samples(samples::DensitySampleVector, fraction::Real =
...
+    n = length(eachindex(v))
+    weights = ones(Int, n)
+
+    i=2
+    k = 0
+    while i <= n-k
+        if v[i] == v[i-1]
+            weights[i-1] += 1
+            deleteat!(v, i)
+            deleteat!(weights, i)
```
oschulz, line 349: `deleteat!` is a very costly operation. Better build the new `v` and `weights` from the ground up, pushing into (initially empty) vectors.

src/samplers/importance/importance_sampler.jl
```diff
@@ -0,0 +1,78 @@
+export SobolSampler, GridSampler, PriorSampler
+
+abstract type ImportanceSampler <: AbstractSamplingAlgorithm end
+
+struct SobolSampler <: ImportanceSampler end
```
oschulz, line 5: `SobolSampler` and `ImportanceSampler` deserve docstrings, and can already be exported and added to "experimental_api.md" as well.

src/samplers/importance/importance_sampler.jl
```diff
@@ -0,0 +1,78 @@
...
+struct GridSampler <: ImportanceSampler end
+
+
+function bat_sample(
+    posterior::AnyPosterior,
+    n::AnyNSamples,
+    algorithm::ImportanceSampler;
+    bounds::Any = var_bounds(posterior)
```
oschulz, line 13: We should not have an additional `bounds` argument here. If the user is unhappy with the automatically derived bounds, they can use `TruncatedDensity` to truncate the prior or posterior. We want to reduce the number of additional arguments to functions like `bat_sample`, to make it easier to nest algorithms (e.g. for the space-partitioned sampling) while keeping the ability to control the behavior of "inner" algorithms.

src/samplers/importance/importance_sampler.jl
```diff
@@ -0,0 +1,78 @@
...
+
+    logvals = density_logval.(Ref(truncated_posterior), samples)
+    weights = exp.(logvals)
+
+    bat_samples = shape.(DensitySampleVector(samples, logval = logvals, weight = weights))
+    return (result = bat_samples, chains = stats)
+end
+
+
+function get_samples(algorithm::SobolSampler, n_samples::Int, posterior::TruncatedDensity)
```
oschulz, line 36: This is just an internal helper function, right? Let's name it `_gen_samples` or so for now.

src/samplers/importance/importance_sampler.jl
```diff
@@ -0,0 +1,78 @@
...
+    dim = length(bounds.bt)
+    ppa = n_samples^(1/dim)
+    ranges = [range(bounds.vol.lo[i], bounds.vol.hi[i], length = trunc(Int, ppa)) for i in 1:dim]
+    p = vec(collect(Iterators.product(ranges...)))
+    return [collect(p[i]) for i in 1:length(p)]
+end
+
+
+
+struct PriorSampler <: AbstractSamplingAlgorithm end
```
oschulz, line 54: Maybe name it `PriorImportanceSampler` or so - would that be a good description?. Also, it deserved a docstring, and can already be exported and added to "experimental_api.md" as well.

src/parameters/density_sample.jl
Cornelius-G: Ah, ok. I actually thought `deleteat` might be less costly than `push` since the maximum length of the array is known and it just gets smaller by deleting some information from the memory, whereas `push` grows to arbitrary length and is therefore not really memory-friendly.

src/samplers/importance/importance_sampler.jl
Cornelius-G: @Lars, will you have a try on renaming and adding a docstring for this?

src/samplers/importance/importance_sampler.jl
Cornelius-G: Lars, will you have a try on adding a docstring for this?

oschulz:
Sorry, you'll need to adapt this to the changes in #141. Was necessary for development of #139 to continue.

```julia
function bat_sample(posterior::AnyPosterior, n::AnyNSamples, algorithm::...
```

needs to become

```julia
function bat_sample_impl(rng::AbstractRNG, posterior::AnyPosterior, n::AnyNSamples, algorithm::...
```

That should be all. The `rng` argument is necessary, even if the `rng` will not be used (it's still optional to the user in `bat_sample`, but `bat_sample_impl` needs to have it).

src/parameters/density_sample.jl
oschulz: `push!` enlarges the array by a factor of two every time it runs out of space, so the array is actually not resized that often. You can use `sizehint!` to pre-alloocate a certain size, but the length of the original sample vector maybe way too much.

`deleteat!` needs to move the whole contents of the array after the deleted element every time, that's why it's a costly operation.

VasylHafych:
Hi @Cornelius-G, @larsroehrig. Can you please let me know about the `TruncatedDensity`  structure that you are planning to implement? Namely, which additional functionality will be available for this object and how will it work? 

I am currently implementing MCMC with space partitioning (#139) and I cannot progress further without having truncated priors. So looking forward to it.

Cornelius-G:
> Hi @Cornelius-G, @larsroehrig. Can you please let me know about the `TruncatedDensity` structure that you are planning to implement? Namely, which additional functionality will be available for this object and how will it work?
> 
> I am currently implementing MCMC with space partitioning (#139) and I cannot progress further without having truncated priors. So looking forward to it.

Hi @VasylHafych you can find the `TruncatedDensity` [here](https://github.com/bat/BAT.jl/blob/5dc1ec7c16721bb25d692bd6914154508838ed0c/src/densities/truncated_density.jl)

It is just a struct of `Density` and `VarVolumeBounds`. For now we do not have any additional functionalities, except that you can now access bounds of the density. If you need to get truncated priors, you can maybe do something similar to [this](https://github.com/bat/BAT.jl/blob/5dc1ec7c16721bb25d692bd6914154508838ed0c/src/samplers/importance/importance_sampler.jl) in lines 17-22. We also have the new (internal) function `estimate_finite_bounds` in [here](https://github.com/bat/BAT.jl/blob/5dc1ec7c16721bb25d692bd6914154508838ed0c/src/densities/distribution_density.jl) in lines 70 and following.

oschulz:
> It is just a struct of Density and VarVolumeBounds. 

As soon as this is merged, we can generalize this further, and add things that may be missing for the space partitioning.

VasylHafych:
> > It is just a struct of Density and VarVolumeBounds.
> 
> As soon as this is merged, we can generalize this further, and add things that may be missing for the space partitioning.

How much time approximately do you expect this to take? Wondering whether I should think of a temporary workaround.

Cornelius-G:
> How much time approximately do you expect this to take? Wondering whether I should think of a temporary workaround.

I hope we can merge very soon. We did now (hopefully) resolve all the requested changes. I will just qucikly need to fix the docs now. Then @oschulz can do a new review and when it is ok for him, it should be ready for merge.

oschulz:
Thanks for all the work on this!

PR #139, opened 2020-06-16
[WIP] Partitioned Sampling

VasylHafych:
Hi @oschulz!  

This will be my PR where I will be merging sampling with space partitioning. I have started with whitening transformation, as you suggested.

Unfortunately, it seems to not be possible to make a clean public function `bat_whiten` which acts on a `DensitySampleVector` **A** and returns whitened `DensitySampleVector` **B** using existing functions from AHMI. The whitening function is written in such a way that it acts on the AHMI-specific data structure — `DataSet{T, I}` which contains split samples — and it transforms only samples vectors, without affecting Log-Likelihood values. It then stores the output of whitening transformation on a separate structure `WhiteningResult`, without returning a whitened density sample vector. Log-Likelihood values are then corrected at the very end of the integration while computing HM estimates on reduced volumes. The function `transform_data!` also does additional operation assigning/determining `suggTargetProb` value, which is an AHMI-specific operation that will not be used in any other application. 

So rewriting this function is not trivial and will require a lot of changes in many other functions/structures (basically redesigning AHMI).


oschulz:
> So rewriting this function is not trivial and will require a lot of changes in many other functions/structures

Ok, we'll defer this for a little while, then.

VasylHafych:
Hi @oschulz! Can you please have a look at the structure of the partitioned sampling that I have just committed?

I have defined a new algorithm

```julia
struct PartitionedSampling{S<:AbstractSamplingAlgorithm,
    I<:IntegrationAlgorithm, P<:SpacePartitioningAlgorithm} <: AbstractSamplingAlgorithm
    samplingalgm::S
    integrationalgm::I
    partitonalgm::P
end
```
with a default ctor 
```julia
function PartitionedSampling()
    PartitionedSampling(MetropolisHastings(), AHMIntegration(), KDBinaryTree())
end
```
and partitioned sampling can be initialized by calling 

```julia
function bat_sample(
        posterior::PosteriorDensity,
        n::Tuple{Integer,Integer},
        algorithm::PartitionedSampling;
        n_subspaces::Integer,
        sampling_kwargs::NamedTuple,
        integration_kwargs::NamedTuple,
        space_part_kwargs::NamedTuple
    )
```
I would appreciate having your code review comments if changes are needed.

VasylHafych:
I already see a discrepancy in the `PartitionedSampling` struct definition. I guess `S<:AbstractSamplingAlgorithm` should be replaced with something else. Right?

oschulz:
#141 should now have solved the problems with dependencies  on abstract algorithm types, when nesting algorithms. All abstract algorithms are now in "src/algotypes".

VasylHafych:
Thanks. Another question:

Space partitioning works only with a flat prior. So I have to transform for example
```julia
prior = NamedTupleDist(x = 0. .. 10., y = 0. .. 10.)
```
into a set of `N` rectangular priors — based on partitioning tree — that cover initial prior: 
```julia
prior_1 = NamedTupleDist(x = 0. .. 5., y = 0. .. 10.)
prior_2 = NamedTupleDist(x = 5. .. 10., y = 0. .. 10.)
```
Do you have ideas on how to do this conveniently, for different prior shapes?

oschulz:
> Space partitioning sampling works only with a flat prior.

We'll have a `TruncatedDensity` soon, as part of #131

VasylHafych:
Hi @Cornelius-G. I will write a reply here to avoid populating your PR with my questions. 

For example, if my task is to split this prior 
```julia

prior =  NamedTupleDist(x = Normal()) # x ∈ [-Inf, Inf]

```
into two parts

```julia
prior_1 =  NamedTupleDist(x = Normal()) # x ∈ [-Inf, 0]

prior_2 =  NamedTupleDist(x = Normal()) # x ∈ [0, Inf]
```

can I do this now with `TruncatedDensity`? If so, will the `Normal()` distribution be renormalized due to truncation?

src/samplers/partitioned_sampling/kd_tree.jl
```diff
@@ -0,0 +1,206 @@
+# This file is a part of BAT.jl, licensed under the MIT License (MIT).
+
+"""
+    partition_space(
+		samples::DensitySampleVector,
+		n_partitions::Integer,
+		algorithm::KDTreePartitioning
+	)
```
oschulz, line 8: Mark that function as BAT-internal

src/samplers/partitioned_sampling/partition_algorithms.jl
```diff
@@ -0,0 +1,49 @@
...
+
+K-D binary space partitioning algorithm. By default, all parameters are
+considered for partitioning. Partition parameters can be specified
+manually by using `partition_dims` argument:
+
+	PartitionedSampling(partition_dims=[1,2,3,4])
+
+"""
+@with_kw struct KDTreePartitioning <: SpacePartitioningAlgorithm
+	partition_dims::Union{Array{Int64,1}, Bool} = false
```
oschulz, line 17: Better use a `Union{AbstractArray{Int64,1}, Missing}`

src/samplers/partitioned_sampling/partition_algorithms.jl
```diff
@@ -0,0 +1,49 @@
...
+end
+
+export KDTreePartitioning
+
+
+"""
+	SpacePartTree
+
```
oschulz, line 25: Mark this as BAT-internal

src/samplers/partitioned_sampling/partition_algorithms.jl
```diff
@@ -0,0 +1,49 @@
...
+	* bounds : Low and high bound of the tree leaf.
+	* left_child : The left child of the tree, `false` is the node is terminal.
+	* right_child : The right child of the tree, `false` is the node is terminal.
+	* cut_axis : Axis along which a cut is performed, `false` if the node is terminal.
+	* cut_coordinate : Coordinate at which a cut is performed, `false` if the node is terminal.
+	* cost : The sum of the cost functions over leaves.
+   * cost_part : The cost functions of the current leaf.
+"""
+mutable struct SpacePartTree
+   terminate::Bool
```
oschulz, line 39: Call this `is_leaf`

src/samplers/partitioned_sampling/partition_sample.jl
```diff
@@ -0,0 +1,102 @@
...
+`AHMIntegration` integrator, and `KDTreePartitioning`:
+
+    PartitionedSampling()
+
+"""
+@with_kw struct PartitionedSampling{S<:AbstractSamplingAlgorithm,
+    I<:IntegrationAlgorithm, P<:SpacePartitioningAlgorithm} <: AbstractSamplingAlgorithm
+    exploration_algm::S = MetropolisHastings()
+    partiton_algm::P = KDTreePartitioning()
+    sampling_algm::S = MetropolisHastings()
```
oschulz, line 19: Call it `sampler`

src/samplers/partitioned_sampling/partition_sample.jl
```diff
@@ -0,0 +1,102 @@
...
+The default constructor is using `MetropolisHastings` sampler,
+`AHMIntegration` integrator, and `KDTreePartitioning`:
+
+    PartitionedSampling()
+
+"""
+@with_kw struct PartitionedSampling{S<:AbstractSamplingAlgorithm,
+    I<:IntegrationAlgorithm, P<:SpacePartitioningAlgorithm} <: AbstractSamplingAlgorithm
+    exploration_algm::S = MetropolisHastings()
+    partiton_algm::P = KDTreePartitioning()
```
oschulz, line 18: Call it `partitioner`

src/samplers/partitioned_sampling/partition_sample.jl
```diff
@@ -0,0 +1,102 @@
...
+
+The default constructor is using `MetropolisHastings` sampler,
+`AHMIntegration` integrator, and `KDTreePartitioning`:
+
+    PartitionedSampling()
+
+"""
+@with_kw struct PartitionedSampling{S<:AbstractSamplingAlgorithm,
+    I<:IntegrationAlgorithm, P<:SpacePartitioningAlgorithm} <: AbstractSamplingAlgorithm
+    exploration_algm::S = MetropolisHastings()
```
oschulz, line 17: Use `exploration_sampler::S = sampler`, should come after `sampler`

src/samplers/partitioned_sampling/partition_sample.jl
```diff
@@ -0,0 +1,102 @@
...
+    partiton_algm::P = KDTreePartitioning()
+    sampling_algm::S = MetropolisHastings()
+    integration_algm::I = AHMIntegration()
```
oschulz, line 20: Call it `integrator`

src/samplers/partitioned_sampling/partition_sample.jl
```diff
@@ -0,0 +1,102 @@
...
+partitioning algorithm specified in `algorithm` with corresponding kwargs
+given by `exploration_kwargs`, and `sampling_kwargs`.
+`n` must be a tuple `(nsteps, nchains, npartitions)`. `posterior` must be a uniform
+distribution for each dimension.
+"""
+function bat_sample(
+    posterior::PosteriorDensity,
+    n::Tuple{Integer,Integer, Integer},
+    algorithm::PartitionedSampling;
+    exploration_kwargs::NamedTuple = NamedTuple(),
```
oschulz, line 45: Put `exploration_kwargs`, `sampling_kwargs` into the algorithm object for now, as a workaround.

src/samplers/partitioned_sampling/partition_sample.jl
```diff
@@ -0,0 +1,102 @@
...
+    sampling_kwargs::NamedTuple = NamedTuple(),
+    n_exploration::Tuple{Integer,Integer} = (10^2, 40)
```
oschulz, line 47: This should go into the algorithm object

src/samplers/partitioned_sampling/partition_sample.jl
```diff
@@ -0,0 +1,102 @@
...
+    integras_subspace = bat_integrate(samples_subspace, integration_algorithm).result
+
+    # α = exp(log(integras_subspace) + log(prior_normalization))
+    α = exp(log(integras_subspace) + log(1.0))
+
+    # samples_subspace.weight .= α.val .* samples_subspace.weight ./ sum(samples_subspace.weight)
+    samples_subspace.weight .= 2 .* samples_subspace.weight # How to change weights to Float?
+
+    info_subspace = TypedTables.Table(
+            integral_loglik = [α],
```
oschulz, line 95: Take `integral_loglik` out

src/samplers/partitioned_sampling/partition_sample.jl
```diff
@@ -0,0 +1,102 @@
...
+            integral_posterior = [integras_subspace],
```
oschulz, line 96: Let's call this `density_integral`

src/samplers/partitioned_sampling/partition_sample.jl
```diff
@@ -0,0 +1,102 @@
...
+            time_mcmc = [1.],
+            time_ahmi = [2.0]
```
oschulz, line 98: Better `integration_time` and `sampling_time`

VasylHafych:
Hi @oschulz, @Cornelius-G. With new changes that we have from #131, I would like to do something like: 

```julia
truncated_posterior = BAT.TruncatedDensity(posterior, manually_given_bounds)

samples = bat_sample(truncated_posterior, (1000, 4), MetropolisHastings()).result
```
Am I understand correctly that this is not yet implemented? If so, do you have any suggestions on how to implement that?

oschulz:
Yes, that still doesn't quite work, but I'll get on it soon.

VasylHafych:
> Yes, that still doesn't quite work, but I'll get on it soon.

OK, In order to sample subspaces, I need to send `TruncatedDensity` which is determined by the partition tree to the `bat_sample` function. I think it is not possible to finalize this PR without this functionality. I would appreciate it if you could fix this, at least for Metropolis sampler.

VasylHafych:
@oschulz by the way, why did you close this PR?

oschulz:
> by the way, why did you close this PR?

Sorry, just clicked the wrong button.

oschulz:
Why do you use `@CPUelapsed` instead of `@elapsed`?

VasylHafych:
@oschulz, I am actually measuring both times. However, the `@elapsed` time is recorded manually, saving the timestamp of the beginning of each evaluation (using `Dates.Time(Dates.now())`) and the timestamp of the end of each (sampling/integration) evaluation. Subtracting them should give us `@elapsed` time. Right? 

I need to store timestamps in order to track how efficiently workers were involved in the sampling process, also to see if there were processes that "stuck" with sampling/integration. This also allows us to plot a nice summary of CPU usage if we run sampling on a cluster (see plot attached). 

I also record `@CPUelapsed` time to see how efficiently threads are involved in sampling/integration. If we plot `@CPUelapsed/@elapsed` we should get a number that is close to `#threads`.

Picture 1

oschulz:
It would be good to record `@elapsed` (or it's equivalent) too, it uses `time_ns()`, which is more precise than `Dates.now()`. For the absolute timestamps you need to use `Dates.now()`, because the offset of `time_ns()` is undefined, so it's not useful with workers on different machines.

`@CPUelapsed` is not without problems, see schmrlng/CPUTime.jl#1. We should compare it to `nthreads() * elapsed`.

VasylHafych:
@oschulz I agree. But it seems to me that we can obtain `@elapsed` time in microseconds by simply subtracting start/end intervals from `Dates.now()`. I think we are not typically interested in measuring sampling time that is in nanoseconds (this is not a typical scale for our problems). But if you think we need this, I can add an additional call of `time_ns()`.

oschulz:
Then leave it as it is for now, maybe. I'm thinking about adding performance measurements for all `bat_...` functions in a generic fashion, and then maybe later on we'll just use that mechanism.

VasylHafych:
@oschulz, I have added a BAT-style plotting recipe that allows plotting a corner plot with a space partition tree. To improve the readability of a partitioning algorithm, it was necessary to plot the upper right corner with interchanged x/y-axes. To implement this, it was necessary to duplicate some code from @Cornelius-G recipe (see the last commit). 

We can now write `plot(samples, tree)` and get: 

download

oschulz:
> To implement this, I have duplicated some code from @Cornelius-G recipe (see the last commit).

Hm, maybe you and Cornelius can figure out a way to factor that code out into a common function, or to make @Cornelius-G's plotting recipe more generic, so that you can use it? It's quite a large chunk of duplicated code ... :-)

Cornelius-G:
@VasylHafych, if it is just that the upper plots need to have x and y axes switched, I guess we could add a keyword to the `upper` and `lower` dictionaries for this. So one could do something like `plot(samples, upper=Dict("swap"=>true)) or similar. Would this be sufficient for your use case?

VasylHafych:
> @VasylHafych, if it is just that the upper plots need to have x and y axes switched, I guess we could add a keyword to the `upper` and `lower` dictionaries for this. So one could do something like `plot(samples, upper=Dict("swap"=>true)) or similar. Would this be sufficient for your use case?

I think it makes sense. I have just modified the recipe, so we can plot the tree by writing 
`plot(outp.result, upper=Dict("partition_tree"=>partition_tree))`
if the partition tree is given, then the upper right corner will be automatically flipped.

oschulz:
You can now use `BAT.truncate_density(posterior, bounds)`, see "test_truncated_density.jl"

src/samplers/partitioned_sampling/partition_sample.jl
```diff
@@ -163,15 +163,13 @@ function convert_to_posterior(posterior::PosteriorDensity, partition_tree::Space
     #Get flattened rectangular parameter bounds from tree
     subspaces_rect_bounds = get_tree_par_bounds(partition_tree)
 
-    #ToDo: Should be a better way to get n_params from posterior:
     n_params = size(subspaces_rect_bounds[1])[1]
     posterior_array = PosteriorDensity[]
```
oschulz, line 167: This won't be type stable. Don't use a for-loop and `push!`, use `map` or `broadcast`.

VasylHafych: [Fixed!](https://github.com/bat/BAT.jl/pull/139/commits/418806a681bb482f1d6ddf03379b69115213772a)

oschulz:
Ok, let me know when it's ready. :-)

VasylHafych:
Hi Oliver

I have been testing sampling with space partitioning today and it is working just perfectly for problems with up to 4 dimensions. But I was not able to get reasonable results even for the normal distribution in 5D (see plot attached). 

I have spent today almost 8 hours trying to figure this out but I was not successful. My ideas: (a) Something has changed in the AHMI; (b) Truncation is not working correctly; (c) AHMI cannot integrate correctly truncated distributions. 

Otherwise, I have finished implementing the space partitioning code and all my tests (1-4D) show perfect performance. So I expect no changes in the code and you can do a review. I have also prepared a tutorial notebook where details of this implementation are described (I can send it to you). 

Would you be available for a call tomorrow to discuss it (could be before or after BAT call)? 

Cheers,
Vasyl

VasylHafych:
Hi All,

I apologize for the spam email. It should have been sent just to Oliver :)

Cheers,
Vasyl

src/BAT.jl
```diff
@@ -58,6 +58,11 @@ using LaTeXStrings
 import Random123
 import TypedTables
 
+#for Space Partitioning
+using CPUTime
+using Dates
+using HypothesisTests
```
oschulz, line 64: Please use `import`, not using here, we want to be able to find use of HypothesisTests in the code later on.

src/BAT.jl
```diff
@@ -58,6 +58,11 @@ using LaTeXStrings
 import Random123
 import TypedTables
 
+#for Space Partitioning
+using CPUTime
```
oschulz, line 62: Please use `import CPUTime` (see below).

src/BAT.jl
```diff
@@ -58,6 +58,11 @@ using LaTeXStrings
...
+using Dates
```
oschulz, line 63: Since Dates is a standard library, please move this up to the `using` statements for the other standard libraries.

src/integration/ahmi_integration.jl
```diff
@@ -85,7 +85,7 @@ function bat_integrate_impl(posterior::DensitySampleVector, algorithm::AHMIntegr
     info = hmi_data.integralestimates
 
     integral = Measurements.measurement(result.estimate, result.uncertainty)
-    (result = integral, info = info)
+    (result = integral, info = hmi_data) # this is needed to debug integration. can be removed later.
```
oschulz, line 88: You can remove `info = ...` above, then.

src/plotting/recipes_ahmi.jl
```diff
@@ -22,7 +22,7 @@ end
 
 
 
-@recipe function f(data::HMIData, dim1::Integer = 1, dim2::Integer = 2; rscale = 0.5,
+@recipe function f(data::HMIData; dim1 = 1, dim2 = 2, rscale = 0.5,
```
oschulz, line 25: kwargs should still have types (e.g. `Integer` and `Bool` here), so that people know what the function expects.

src/plotting/recipes_samples_2D.jl
```diff
@@ -58,14 +58,15 @@
         acc = findall(x -> x > 0, samples.weight)
         rej = findall(x -> x <= 0, samples.weight)
 
-        color = parse(RGBA{Float64}, get(plotattributes, :seriescolor, :green))
+        color = parse(RGBA{Float64}, get(plotattributes, :seriescolor, :blue))
```
oschulz, line 61: @Cornelius-G , can you comment on these changes? This is your code, is this fine with you?

src/plotting/recipes_samples_2D.jl
```diff
@@ -58,14 +58,15 @@
...
         label = get(plotattributes, :label, isempty(rej) ? "samples" : "accepted")
 
         @series begin
             seriestype := :scatter
             label := label
-            markersize := [w < 1 ? base_markersize : base_markersize * sqrt(w) for w in samples.weight[acc]]
+            markersize := [w < 1 ? base_markersize : base_markersize * 0.5*log(w) for w in samples.weight[acc]]
```
oschulz, line 67: @Cornelius-G , can you comment on these changes? This is your code, is this fine with you?

src/plotting/recipes_samples_overview.jl
```diff
@@ -32,6 +32,9 @@
     layout --> nparams^2
     size --> (1000, 600)
 
+    partition_tree = get(upper, "partition_tree", false)
```
oschulz, line 35: @Cornelius-G , can you comment on these changes (and below)? This is your code, is this fine with you?

src/plotting/recipes_samples_overview.jl
```diff
@@ -187,3 +224,10 @@ end
         end
     end
 end
+
+
+function rect_box(x_s::F, x_f::F, y_s::F, y_f::F) where {F<:AbstractFloat}
```
oschulz, line 229: If this is purely an internal helper function, maybe call it `_rect_box`

oschulz:
I'm asking @Cornelius-G to review this too, since it involves changes to the plotting code.

src/plotting/recipes_samples_2D.jl
Cornelius-G: @VasylHafych why was this change necessary? But I guess this should be fine.

src/plotting/recipes_samples_overview.jl
Cornelius-G: This is also fine for me. But @VasylHafych is there also a recipe to do something like:
```
plot(samples, (:a, :b))
plot!(partition_tree, (:a, :b))
```
?
I guess it would be good to have this additionally so the user can also plot this partition boxes on top of their (potentially customized) plots.
These `plot(samples)` plots are just intended for some kind of overview and cannot be customized as good as the other 1D and 2D plots.

src/plotting/recipes_samples_2D.jl
Cornelius-G: Well this is some kind of personal preference, I guess.
This code for scatter plots is actually from @oschulz. I guess red for rejected and green for accepted is a bit more natural choice.

oschulz: Oh, right - Vasyl, let's stick with green for accepted and red for rejected, should be immediately intuitive to the user. Blue can then be reserved for reference samples or similar.

VasylHafych: Hi all. Actually, I was not planning to include this color-change into the final PR. I have adjusted colors to be able to plot exploration samples, tree, and histogram together. So this was suiting the best for my problem. But I will leave the original version in the final PR.

src/plotting/recipes_samples_2D.jl
VasylHafych: When I plot exploration samples (using :scatter) I need them to be in a smaller size to see more clearly their density. For now, I have just decreased marker size manually before we make this option adjustable using `kwargs`.

src/plotting/recipes_samples_overview.jl
VasylHafych: Great suggestion, thanks! I will add this function when the algorithm is tested.

src/plotting/recipes_ahmi.jl
VasylHafych: Julia is complaining that this is not possible, that's why I have removed it: 

```julia
┌ Info: Precompiling BAT [c0cd4b16-88b7-57fa-983b-ab80aecada7e]
└ @ Base loading.jl:1260
┌ Warning: Type annotations on keyword arguments not currently supported in recipes. Type information has been discarded
└ @ RecipesBase ~/.julia/packages/RecipesBase/jcXIg/src/RecipesBase.jl:116
┌ Warning: Type annotations on keyword arguments not currently supported in recipes. Type information has been discarded
└ @ RecipesBase ~/.julia/packages/RecipesBase/jcXIg/src/RecipesBase.jl:116

```

src/integration/ahmi_integration.jl
VasylHafych: Done!

src/plotting/recipes_ahmi.jl
oschulz: Oh, of course, it's a recipe!

src/BAT.jl
VasylHafych: Done!

src/plotting/recipes_samples_2D.jl
Cornelius-G: Ah, ok. Good point, this should be adjustable with keywords.

src/plotting/recipes_samples_overview.jl
VasylHafych: Sure, done!

VasylHafych:
Thanks! Ready for the next round of comments :)

src/plotting/recipes_samples_2D.jl
```diff
@@ -64,8 +64,9 @@
...
```
oschulz, line 67: I'm not so happy with this - the area of the marker should be proportional to the weight

VasylHafych:
@oschulz, small summary: I was able to run the algorithm using Normal distribution and Gaussian Shell (from 1 up to 18 dimensions). Cauchy distribution is working only up to 10 dimensions.

VasylHafych:
Also, I tried running 
```julia
bat_eff_sample_size(samples[binsry_mask])
```
But this is not working. I need this to estimate the effective number of samples on each subspace which is given by `binsry_mask` indices. Could you suggest how to fix this?

oschulz:
Try `bat_eff_sample_size(unshaped.(samples[binsry_mask]))`.

`bat_eff_sample_size` doesn't work on shaped samples (yet).

oschulz:
> I was able to run the algorithm using Normal distribution and Gaussian Shell (from 1 up to 18 dimensions). Cauchy distribution is working only up to 10 dimensions.

Well that's not bad at all, right? :-)

VasylHafych:
Hi @oschulz! Yes, I have tested sampling using MPCDF machines and performance scales with the number of workers very well. I think this is now ready to be merged.

src/plotting/recipes_samples_2D.jl
VasylHafych: done!

oschulz:
Thanks for all your hard work! I'll merge this now, we can also do more fine-tuning later on.

PR #145, opened 2020-06-24
[WIP] Performance Test Suite

salvolc:
So far running the scripts in example/run_benchmark(1 or 2)D.jl will give plots and some statistics of the performance.
Statistics include the modes, mean, variation and chi2 (chi2 only in the 1D case) and plotting the distributions, differences and pull plots.
Multi modal Cauchy and Funnel are used from the implemented BAT test functions.
Exp, Gaussian and Cauchy are taken from Distributions.jl.
Rastrigin, SineSquared and HoelderTable functions implemented as densities. 
GaussianShell so far included by hand.


oschulz:
I don't think we should commit the run-stats to the repo, they're too system-dependent.

salvolc:
Yeah, definitely. Should be removed now.

oschulz:
Is this still WIP, of finished already?

salvolc:
Kind of depends. I would argue the performance testing in general is finished.
The only things worth discussing is for one implementing the gaussian shell (however this can be done rather quick when the distribution is available) and the test densities. The implementation of them generally work but they only give normalized posteriors when used with uniform priors, however the bounds can be chosen freely except for this HoelderTable one (because it doesn't have an analytical solution for the integral). So some sort of constrain should be implemented for that case. Otherwise I would consider this pretty much finished.

oschulz:
Ok, then I suggest we take the gaussian shell out, and do that as a separate PR.

oschulz:
It looks like there's some duplication between #145 and #112 - @salvolc, @tldrma, can you merge your efforts, here?

salvolc:
I'll take the gaussian shell out for now.
Yeah, I have merged @tldrma fork to mine to implement the test functions, I don't know what way would be the best to 'merge' the two pull requests. I could take the diffs out of this fork but then the performance won't run until both forks merged into BAT.jl then.

tldrma:
I have an implementation for gaussian shell that works with rand(), but I haven't gotten it to work with bat_sample() yet. I started PR for it if anyone wants to see it.

oschulz:
Ok, then let's merge #112 first (Scott, is that ready to be merged), then you can rebase #145 on top of the new master and remove duplications in #145 after (and possibly improve what #112 added on the way).

salvolc:
Sounds good to me.

tldrma:
@oschulz It should be ready, as long as you have no other changes requested.

examples/benchmarks/functions_1D.jl
```diff
@@ -0,0 +1,224 @@
+########1D functions for benchmarking###########
+################################################
+name_normal = "normal"
```
oschulz, line 3: Better make a Dict of test functions, with `name => f`, `f()` should then create the function/distribution.

examples/benchmarks/functions_1D.jl
```diff
@@ -0,0 +1,224 @@
+########1D functions for benchmarking###########
+################################################
+name_normal = "normal"
+analytical_integral_normal = 1
```
oschulz, line 4: All integrals should be one, implicitly (otherwise please change functions/distributions accordingly). These `analytical_integral_...` constants should be removed.

examples/benchmarks/functions_1D.jl
```diff
@@ -0,0 +1,224 @@
+########1D functions for benchmarking###########
+################################################
+name_normal = "normal"
+analytical_integral_normal = 1
+
+normal =
+	params -> begin
+	   	return LogDVal(logpdf(Normal(0,1),params.x))
+	end
+prior_normal = NamedTupleDist(
```
oschulz, line 10: We don't want to restrict distributions with infinite support here - if necessary, the sampler will have to find proper bounds.

examples/benchmarks/functions_1D.jl
```diff
@@ -0,0 +1,224 @@
...
+normal =
+	params -> begin
+	   	return LogDVal(logpdf(Normal(0,1),params.x))
+	end
+prior_normal = NamedTupleDist(
+	x = -7..7
+)
+posterior_normal = PosteriorDensity(normal,prior_normal)
+analytical_function_normal = x -> pdf(Normal(0,1),x)
+analytical_stats_normal = [0,0,1]#
```
oschulz, line 15: Which stats are these? If they can be retrieved by mean, var, etc. on the distribution, we don't need consts here.

examples/benchmarks/functions_1D.jl
```diff
@@ -0,0 +1,224 @@
...
+posterior_cauchy = PosteriorDensity(cauchy,prior_cauchy)
+analytical_function_cauchy = x -> pdf(Cauchy(0,1),x)
+analytical_stats_cauchy = [0,0,15.2459]#(50-2*atan(25))/pi] night ganz
+################################################
+
+################################################
+name_multi_cauchy = "multi cauchy"
+analytical_integral_multi_cauchy = 0.029053426970
+
+multi_cauchy = let a=5., b=4.
```
oschulz, line 38: This should be enough: `multi_cauchy_testdist() = Product([Cauchy(-1*a, b), Cauchy(a, b)])`

examples/benchmarks/functions_1D.jl
```diff
@@ -0,0 +1,224 @@
...
+analytical_stats_multi_cauchy_variance = (2*(80 + 21/2*atan(35/2) + 21/2*atan(45/2) - 5*log(2029/1229)))/π
+analytical_stats_multi_cauchy = [[-5,5],0,analytical_stats_multi_cauchy_variance]
+#Analytical Variance (2*(80 + 21/2*atan(35/2) + 21/2*atan(45/2) - 5*log(2029/1229)))/π
+################################################
+
+################rastrigin x-cos(x)##############
+name_rastrigin = "rastrigin"
+analytical_integral_rastrigin = 1
+
+rastrigin =
```
oschulz, line 58: See above.

examples/benchmarks/functions_1D.jl
```diff
@@ -0,0 +1,224 @@
...
+################################################
+name_hoelder = "hoelder table"
+analytical_integral_hoelder = 1
+
+hoelder = let l=1
+    params -> begin
+		return LogDVal(BAT.density_logval(HoelderTable(),params.x))
+	end
+end
+prior_hoelder = NamedTupleDist(
```
oschulz, line 115: If a specific prior is needed, let's do this: Define a `get_test_prior(d::SomeDensity) = ...` that can be used in a generic fashion.

examples/benchmarks/run_benchmark_1D.jl
```diff
@@ -0,0 +1,23 @@
...
+
+algorithm = MetropolisHastings()
+n_chains = 8
+n_samples = 10^5
+
+for i in 1:length(name)
+    sample_stats_all = run1D(
+        posterior[i],
+        name[i],
+        analytical_integral[i],
```
oschulz, line 15: This should always be 1

examples/benchmarks/run_benchmark_1D.jl
```diff
@@ -0,0 +1,23 @@
...
+        func[i],
+        analytical_stats[i],
```
oschulz, line 17: Mean, cov, etc. should be defined for the Densities/Distributions already.

examples/benchmarks/run_benchmark_1D.jl
```diff
@@ -0,0 +1,23 @@
+using BAT, ValueShapes, IntervalSets, Distributions, Plots
+using StatsBase, ArraysOfArrays, LinearAlgebra, LaTeXStrings, QuadGK, PrettyTables
+
+include("utils.jl")
+include("functions_1D.jl")
+
```
oschulz, line 6: This should be wrapped into a function

src/densities/testDensities.jl
```diff
@@ -0,0 +1,59 @@
...
+
+function Rastrigin(;bounds=[-3.,3.],a=30,b=10)
+    integral_val = -a*bounds[1] + b*bounds[1] + (bounds[1]^3)/3 + a*bounds[2] - b*bounds[2] - (bounds[2]^3)/3 + (b*(-sin(2*bounds[1]*pi) + sin(2*bounds[2]*pi)))/(2*pi)
+    dist = x-> (1/integral_val) * (a - b - x^2 + b*cos(2*pi*x))
+    Rastrigin(bounds,a,b,dist)
+end
+
+BAT.density_logval(density::Rastrigin,v::Any) = log(broadcast(density.dist,v))
+
+"""
```
oschulz, line 18: The doc-string needs to come directly before the struct, no whitespace in between.

src/densities/testDensities.jl
```diff
@@ -0,0 +1,59 @@
...
+    bounds = [-10.,10.]
+    integral_val = (-18*exp(1)^2*π^2 - 12*exp(1)^3*π^2 + exp(1)*(500 + 488*π^2) - 6*exp(1)^(10/π)*π*(π*cos(10) - sin(10)))/(3*exp(1)*(1 + π^2))
+    dist = x-> (1/integral_val) * (10-((x^2)/20) - abs( sin(x)*exp(abs(1-((sqrt(x^2))/(pi))))) )
+    HoelderTable(bounds,dist)
+end
+
+BAT.density_logval(density::HoelderTable,v::Any) = log(broadcast(density.dist,v))
+
+
+"""
```
oschulz, line 56: See above-

src/densities/testDensities.jl
```diff
@@ -0,0 +1,59 @@
...
+    SineSquared <: AbstractDensity
+"""
+
+struct HoelderTable{B<:Vector{Float64}} <: AbstractDensity
+    bounds::B
+    dist
+end
+export HoelderTable
```
oschulz, line 44: Don't export

src/distributions/funnel.jl
```diff
@@ -0,0 +1,86 @@
+struct Funnel{T<:Real, V<:AbstractVector, dimensions<:Int64, L<:ContinuousMultivariateDistribution} <: ContinuousMultivariateDistribution
```
oschulz, line 1: Should have a docstring

src/distributions/funnel.jl
```diff
@@ -0,0 +1,86 @@
...
+end
+
+Base.length(d::Funnel) = d.n
+Base.eltype(d::Funnel) = eltype(d.λ)
+
+Distributions.mean(d::Funnel) = Distributions.mean(d.likelihood)
+Distributions.var(d::Funnel) = Distributions.var(d.likelihood)
+Distributions.std(d::Funnel) = Distributions.std(d.likelihood)
+
+## Maybe this is incorrect, because x should be a variable?
```
oschulz, line 25: We should check if it's correct or not before merging. :-)

src/distributions/multimodal_cauchy.jl
```diff
@@ -0,0 +1,53 @@
+struct MultimodalCauchy{T<:AbstractVector, MM<:MixtureModel, P<:Product}  <: ContinuousMultivariateDistribution
+    dists::T
+    MM1::MM
+    MM2::MM
+    n::Int64
+    likelihood::P
+end
+
+"""
```
oschulz, line 9: See above.

oschulz:
> It should be ready, as long as you have no other changes requested.

Still needs a little work. :-)

src/distributions/funnel.jl
tldrma: I think this is the older version of funnel.jl? I don't call it likelihood anymore. It should just have parameters a, b, n.

I also just have Distributions.mean(d::FunnelDistribution) = zeros(d.n) and didn't use var/std, because of not knowing them a priori.

oschulz: Ok, just take such commented-out code out, less confusion. :-)

oschulz:
Ok, you can rebase this on #112 now. :-)

oschulz:
#112 is now on master.

tldrma:
I saw that on one of the travis builds that the multimodal cauchy integral test failed. It seems like it's pretty uncommon from when I ran the test a lot. Should I do a quick pull request to increase the error tolerance?

oschulz:
> Should I do a quick pull request to increase the error tolerance?

Maybe a bit, yes, sporadic failures because of that are a nuisance. We shouldn't be too strict with these tolerances for Travis testing.

salvolc:
The rebase should be done, not exactly sure if everything went right but the commits seem to be ok.
I will work on the improvements now :)

oschulz:
Well, the main thing should be that you can use the multimodal-cauchy and the funnel-distribution that @tldrma added now.

examples/benchmarks/functions_1D.jl
```diff
@@ -0,0 +1,224 @@
...
+analytical_stats_exp1 = [0,1,1]
+################################################
+
+################################################
+name_hoelder = "hoelder table"
+analytical_integral_hoelder = 1
+
+hoelder = let l=1
+    params -> begin
+		return LogDVal(BAT.logvalof_unchecked(HoelderTable(),params.x))
```
oschulz, line 112: Normally, code should always call `logvalof`, not `BAT.logvalof_unchecked` - that's only for *implementing* a density, not for calling it. Are there any problems here using `logvalof`? If so, that's probably an issue we need to fix.

src/densities/testDensities.jl
```diff
@@ -0,0 +1,57 @@
...
+end
+
+
+function Rastrigin(;bounds=[-3.,3.],a=30,b=10)
+    integral_val = -a*bounds[1] + b*bounds[1] + (bounds[1]^3)/3 + a*bounds[2] - b*bounds[2] - (bounds[2]^3)/3 + (b*(-sin(2*bounds[1]*pi) + sin(2*bounds[2]*pi)))/(2*pi)
+    dist = x-> (1/integral_val) * (a - b - x^2 + b*cos(2*pi*x))
+    Rastrigin(bounds,a,b,dist)
+end
+
+BAT.logvalof_unchecked(density::Rastrigin,v::Any) = log(broadcast(density.dist,v))
```
oschulz, line 15: Normally, code should always call `logvalof`, not `BAT.logvalof_unchecked` - that's only for *implementing* a density, not for calling it. Are there any problems here using `logvalof`? If so, that's probably an issue we need to fix.

examples/benchmarks/functions_1D.jl
salvolc: Yes it gives errors. Using `BAT.logvalof` instead of `BAT.logvalof_unchecked` gives an 
`Density evaluation failed at v = (x = 0.5032303096189046,) due to exception MethodError(ValueShapes.unshaped, (0.5032303096189046,), 0x0000000000006de9), density has type BAT.GenericDensity{var"#169#170"}`
Error, I tried that first but then just tried `_unchecked` and it worked to I didn't investigate further.

Cornelius-G: @salvolc did you use the latest master branch? Had similar issues here #156, which were fixed by https://github.com/bat/BAT.jl/commit/916526377f321f279a217be9ca1597e9755c1412

salvolc: Yes, I am one commit behind the master right now so at 088406083a61dfe99a998f7fc3d621c73dbc16dd which, however, should be after that bugfix.

oschulz: Ah, I think I know what it is. I'll try to fix that via ValueShapes.jl, will take a few hours to get to you, needs a new release.

examples/benchmarks/functions_1D.jl
salvolc: These are analytical stats of the distributions, one could for some distributions use the input for the mean or var used to create the distribution in the first place. But other are way more complex, like the expected variance of the cauchy distribution within bounds (or the multi-modal cauchy even) or the mean distribution of a sin^2 function. While these do depend on the paramaters used for the distribution they cannot easily be calculated unless using some symbolic for instance.
For the implementation inside the code there definitely be a smarter way of handling these stats.

examples/benchmarks/functions_1D.jl
```diff
@@ -0,0 +1,224 @@
+########1D functions for benchmarking###########
+################################################
+name_normal = "normal"
+analytical_integral_normal = 1
+
+normal =
```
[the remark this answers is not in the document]
salvolc: It kind of depends, if I don't implement a log-posterior function there would be no way to ensure the normalization of some distributions. For instance the multimodal cauchy doesn't integrate to one, so if you want to compare the sample to the true likelihood one would need to integrate numerically somewhere else to have the same ranges for both plots. This is also the reason I always chose flat priors with bounds. For example the cauchy distribution doesn't integrate perfectly to one within -40 to 40, implementing a normalization factor allows to still compare it to its analytical form and having the finite bounds makes it possible to precalculate the stats (since the integration to calculate the var and mean depend on the bound of the functions) and in that case particularly the integral for the var doesn't converge for -inf to inf.

examples/benchmarks/functions_1D.jl
oschulz: @salvolc try now, based on the current BAT master

salvolc: @oschulz yes, it's working.

examples/benchmarks/functions_1D.jl
oschulz: Ah, right. Then maybe make a `NamedTuple` like `(posterior = ..., pdf = ..., var = ...)` for each test case, and put them in a single `Dict` that contains all test cases.

examples/benchmarks/functions_1D.jl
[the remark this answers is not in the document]
oschulz: Ah, right, because of the bounds. Ok, then add that to the `NamedTuple` for each test case (see other comment).

examples/benchmarks/functions_1D.jl
salvolc: Yeah, sounds like a good suggestion, I'll try it out.

salvolc:
@tldrma @oschulz It seems that using the BAT Distributions for sampling with Metropolis Hasting doensn't work.
`bat_sample(Normal(0,1), (10^5,4), MetropolisHastings())` This works fine 
`bat_sample(BAT.FunnelDistribution(1.2,0.,1), (10^5,4), MetropolisHastings())` This results in 
`MethodError: no method matching cov(::BAT.FunnelDistribution)` any idea what went wrong?

tldrma:
> @tldrma @oschulz It seems that using the BAT Distributions for sampling with Metropolis Hasting doensn't work.
> `bat_sample(Normal(0,1), (10^5,4), MetropolisHastings())` This works fine
> `bat_sample(BAT.FunnelDistribution(1.2,0.,1), (10^5,4), MetropolisHastings())` This results in
> `MethodError: no method matching cov(::BAT.FunnelDistribution)` any idea what went wrong?

Looks like it's just because there's no function implemented for the covariance matrix `cov` of the funnel distribution. I didn't put it in because I wasn't sure how it's possible with the variance changing every sample.

oschulz:
Just add a numerical cov-estimate for now, like we use for the `HierarchicalDensity`:

```julia
function Statistics.cov(dist::BAT.FunnelDistribution)
    cov(nestedview(rand(bat_determ_rng(), sampler(dist), 10^5)))
end
```

The need to have a covariance will go away soon, I'm building an new system to initialized samplers that will do this automatically.

salvolc:
There is another problem with the test distribution but this time regarding AHMI integration with the Metropolis Hastings samples.
`samples  = bat_sample(BAT.FunnelDistribution(n=1), (n_samples, n_chains), MetropolisHastings()).result`
`bat_integrate(samples,AHMIntegration()).result.val` this will fail due to 
`MethodError: no method matching BAT.DataSet(::Array{Float32,2}, ::Array{Float32,1}, ::Array{Float32,1}, ::Array{Int64,1}, ::Array{Int64,1}, ::Int64, ::Int64, ::Int64, ::Float64, ::Bool, ::Bool, ::BAT.SpacePartitioningTree{Float32,Int64}, ::Array{Int64,1}, ::Float32)`
while
`samples  = bat_sample(BAT.FunnelDistribution(n=1), (n_samples*n_chains)).result`
`bat_integrate(samples,AHMIntegration()).result.val` will give a reasonable estimate for the integral

(also sidenote when not specifying the number of dimensions for BAT.FunnelDistribution() the MetropolisHastings() one failes while sampling due to the Loglikelihood being NaN)

oschulz:
Looks like `BAT.DataSet`, resp. it's ctor, can't handle `Float32`-data properly (but it should!).

@VasylHafych , can you maybe give `BAT.DataSet` a quick overhaul? The `I<:Integer` type parameter can go, just replace it by `Int`. And things like `subsetsize` and `tolerance` can probably be fixed to `Float64`.

VasylHafych:
> (also sidenote when not specifying the number of dimensions for BAT.FunnelDistribution() the MetropolisHastings() one failes while sampling due to the Loglikelihood being NaN)

@sthayashi, I had the same issue. I was not able to run:
```julia
using Statistics, ArraysOfArrays, BAT

function Statistics.cov(dist::BAT.FunnelDistribution)
    cov(nestedview(rand(BAT.bat_determ_rng(), BAT.sampler(dist), 10^5)))
end

samples = bat_sample(FunnelDistribution(n=2), (10^4, 2), MetropolisHastings()).result
```
the error message is: 
```terminal
ERROR: TaskFailedException:
Log-density must not evaluate to NaN, v = Float32[-1157.7933, 2095.2954] , density has type BAT.DistributionDensity{BAT.FunnelDistribution,BAT.HyperRectBounds{Float32}}
Stacktrace:
 [1] _check_density_logval at /Users/vhafych/MPP-Server/gitrepos/BAT.jl/src/densities/abstract_density.jl:169 [inlined]
 [2
```

oschulz:
Thanks - let me know when this is ready to be merged.

salvolc:
Well I am adjusting the higher dimensional tests but there are still a few problems with the test distributions.
`bat_sample(BAT.MultimodalCauchy(),(10^5,8),MetropolisHastings())` doesn't seem to work (also not in 1D) with 

``` 
Failed to generate 64 viable MCMC chains
in top-level scope at run_benchmark_ND.jl:71
in bat_sample at dev/BAT/src/algotypes/sampling_algorithm.jl:110 ...
```
and the funnel distribution seems to fail at higher dimensions as Vasyl mentioned.

oschulz:
Ok, we'll hold off, then

oschulz:
>  I am not exactly sure what causes the error for the multimodal cauchy and funnel distributions

Can you give me a code snippet and error message?

salvolc:
Basically the same as above or do you need more infomation
`bat_sample(BAT.MultimodalCauchy(),(10^5,8),MetropolisHastings())` this willl result in 
```
Failed to generate 64 viable MCMC chains
in top-level scope at test.jl:445
in bat_sample at dev/BAT/src/algotypes/sampling_algorithm.jl:110
in #bat_sample#78 at dev/BAT/src/algotypes/sampling_algorithm.jl:111 
in bat_sample at dev/BAT/src/algotypes/sampling_algorithm.jl:97 
in #bat_sample#76 at dev/BAT/src/algotypes/sampling_algorithm.jl:97 
in bat_sample_impl at dev/BAT/src/samplers/mcmc/mcmc_sample.jl:182 
in #bat_sample_impl#171 at dev/BAT/src/samplers/mcmc/mcmc_sample.jl:183 
in bat_sample_impl at dev/BAT/src/samplers/mcmc/mcmc_sample.jl:139
in #bat_sample_impl#170 at dev/BAT/src/samplers/mcmc/mcmc_sample.jl:145
in mcmc_sample##kw at dev/BAT/src/samplers/mcmc/mcmc_sample.jl:46 
in #mcmc_sample#166 at dev/BAT/src/samplers/mcmc/mcmc_sample.jl:50 
in mcmc_init at dev/BAT/src/samplers/mcmc/mcmc_tuner.jl:257
in error at base/error.jl:33
```
for any amount of dimensions and with that number of failed chaines interestingly always being 8 times the number of chaines given into the code.
The Funnel Distribution failes as soon as dim > 1 like in 
`bat_sample(BAT.FunnelDistribution(n=2),(10^5,8),MetropolisHastings())` it will give 
```
Log-density must not evaluate to NaN, v = Float32[-2739.3545, -7166.9805] , density has type BAT.DistributionDensity{BAT.FunnelDistribution,BAT.HyperRectBounds{Float32}}
Stacktrace:
 [1] _check_density_logval a....
```
dim = 1 works fine.
And for both the iid sampling works like a charm.

oschulz:
Try

    bat_sample(BAT.MultimodalCauchy(),(10^5,8),MetropolisHastings())

and

    bat_sample(BAT.FunnelDistribution(n=2),(10^5,8),MetropolisHastings())

now, with the current BAT master and the latest release of ValueShapes. Not sure why  MultimodalCauchy failed, but it works now. The implementation of `BAT.FunnelDistribution` was less than optimal, that also seems to work now (and `cov` is also defined for the funnel now).

salvolc:
The funnel distribution works, the cauchy didn't on my pc but on a fresh installation on the cluster it seems to be working so things should be fine, thanks.

oschulz:
Gaussian shell should now be fully available (in case it's needed here): #168

oschulz:
@salvolc tests pass now, Ok if I merge this?

salvolc:
Yes, should be good, I will do some work like including the gaussian shell and ESS for ahmi and mh but I will do that after my holiday so I can just do another pr that.

oschulz:
Thanks - done.

PR #147, opened 2020-06-25
[WIP] Gaussian shell rejection sampling

tldrma:
Currently works with rand(), but not bat_sample()

### DIFF
--- a/src/distributions/gaussian_shell.jl
+++ b/src/distributions/gaussian_shell.jl
@@ -1,6 +1,5 @@
 using SpecialFunctions
-using Random123
-
+using Random123 
 import AdaptiveRejectionSampling
 
 """
@@ -12,6 +11,10 @@ Gaussian Shell (Caldwell et al.)[https://arxiv.org/abs/1808.08051].
 - `r::Real`: The radius of the Gaussian shell distribution.
 - `w::Real`: Variance of the Gaussian shell distribution.
 - `n::Int`: Number of dimensions.
+
+Constructors:
+```julia
+BAT.GaussianShell(r::Real, w::Real, c::Vector, n::Int)
 """
 struct GaussianShell{T<:Real, V<:AbstractVector} <: ContinuousMultivariateDistribution
     r::T
@@ -50,9 +53,11 @@ function Distributions._rand!(rng::AbstractRNG, d::GaussianShell, X::AbstractMat
 end
 
 function Distributions._logpdf(d::GaussianShell, x::AbstractArray)
-    normalization = -log(sqrt(2*π*(d.w)^2))
+    pdf_normalization = -log(sqrt(2*π*(d.w)^2))
+    ρ = LinearAlgebra.norm(x)
+    geometric_correction = log(nball_surf_area(ρ, d.n))
     integral_result = -(sqrt(sum((x .- d.c).^2)) - d.r)^2 / (2*d.w^2)
-    result = integral_result + normalization
+    result = integral_result + pdf_normalization - geometric_correction
     return result
 end
 
--- a/src/distributions/gaussian_shell.jl
+++ b/src/distributions/gaussian_shell.jl
@@ -1,0 +1,63 @@
+using SpecialFunctions
+using Random123
+
+import AdaptiveRejectionSampling
+
+"""
+    BAT.GaussianShell([r=5, w=2, n=2])
+
+Gaussian Shell (Caldwell et al.)[https://arxiv.org/abs/1808.08051].
+
+# Arguments
+- `r::Real`: The radius of the Gaussian shell distribution.
+- `w::Real`: Variance of the Gaussian shell distribution.
+- `n::Int`: Number of dimensions.
+"""
+struct GaussianShell{T<:Real, V<:AbstractVector} <: ContinuousMultivariateDistribution
+    r::T
+    w::T
+    c::V
+    n::Int
+end
+function GaussianShell(;r::Real=5, w::Real=2, n::Integer=2)
+    c = zeros(n)
+    r,w = promote(r, w)
+    GaussianShell(r, w, c, n)
+end
+
+nball_surf_area(r, ndims) = 2 * π^(ndims/2) / gamma(ndims/2) * r^(ndims-1)
+gs_pdf_r(r, ndims, r0, w) = nball_surf_area(r, ndims) * 1/(2π * w^2) * exp(-(abs(r) - r0)^2 / (2 * w^2))
+
+function gauss_shell_radial_samples(rng::AbstractRNG, ndims::Integer, r0::Real, w::Real, n::Integer)
+    f(r) = gs_pdf_r(r, ndims, r0, w)
+    sampler = AdaptiveRejectionSampling.RejectionSampler(f, (10^-10, Inf))
+    AdaptiveRejectionSampling.run_sampler!(rng, sampler, n)
+end
+
+function rand_nball_surf_samples(rng::AbstractRNG, ndims::Integer, n::Integer)
+    A = randn(rng, ndims, n)
+    A ./= sqrt.(sum(A .* A, dims = 1))
+end
+
+function Distributions.rand(rng::AbstractRNG, d::GaussianShell, n::Int)
+    Distributions._rand!(rng, d, Matrix{eltype(d)}(undef, length(d), n), n)
+end
+
+function Distributions._rand!(rng::AbstractRNG, d::GaussianShell, X::AbstractMatrix, n::Int)
+    X = rand_nball_surf_samples(rng, d.n, n)
+    R = gauss_shell_radial_samples(rng, d.n, d.r, d.w, n)
+    X .* R'
+end
+
+function Distributions._logpdf(d::GaussianShell, x::AbstractArray)
+    normalization = -log(sqrt(2*π*(d.w)^2))
+    integral_result = -(sqrt(sum((x .- d.c).^2)) - d.r)^2 / (2*d.w^2)
+    result = integral_result + normalization
+    return result
+end
+
+Base.length(d::GaussianShell) = length(d.c)
+
+Base.eltype(d::GaussianShell) = eltype(d.c)
+
+Distributions.params(d::GaussianShell) = (d.r, d.w, d.c)
--- a/src/distributions/gaussian_shell.jl
+++ b/src/distributions/gaussian_shell.jl
@@ -17,7 +17,7 @@ Constructors:
 ```julia
 BAT.GaussianShell(r::Real, w::Real, c::Vector, n::Int)
 """
-struct GaussianShell{T<:Real, V<:AbstractVector, F<:Float64} <: ContinuousMultivariateDistribution
+struct GaussianShell{T<:Real, V<:AbstractVector{<:Real}, F<:AbstractFloat} <: ContinuousMultivariateDistribution
     r::T
     w::T
     c::V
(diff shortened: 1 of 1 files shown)


oschulz:
Thanks, I'll take a look at this.

src/distributions/gaussian_shell.jl
```diff
@@ -0,0 +1,63 @@
...
+    sampler = AdaptiveRejectionSampling.RejectionSampler(f, (10^-10, Inf))
+    AdaptiveRejectionSampling.run_sampler!(rng, sampler, n)
+end
+
+function rand_nball_surf_samples(rng::AbstractRNG, ndims::Integer, n::Integer)
+    A = randn(rng, ndims, n)
+    A ./= sqrt.(sum(A .* A, dims = 1))
+end
+
+function Distributions.rand(rng::AbstractRNG, d::GaussianShell, n::Int)
```
tldrma, line 42: Integer not specific enough

src/distributions/gaussian_shell.jl
```diff
@@ -0,0 +1,63 @@
...
+end
+
+function Distributions._rand!(rng::AbstractRNG, d::GaussianShell, X::AbstractMatrix, n::Int)
+    X = rand_nball_surf_samples(rng, d.n, n)
+    R = gauss_shell_radial_samples(rng, d.n, d.r, d.w, n)
+    X .* R'
+end
+
+function Distributions._logpdf(d::GaussianShell, x::AbstractArray)
+    normalization = -log(sqrt(2*π*(d.w)^2))
```
tldrma, line 53: Not normalizing to 1

src/distributions/gaussian_shell.jl
tldrma, line 42: Integer not specific enough

src/distributions/gaussian_shell.jl
```diff
@@ -50,9 +53,11 @@ function Distributions._rand!(rng::AbstractRNG, d::GaussianShell, X::AbstractMat
 end
 
 function Distributions._logpdf(d::GaussianShell, x::AbstractArray)
-    normalization = -log(sqrt(2*π*(d.w)^2))
+    pdf_normalization = -log(sqrt(2*π*(d.w)^2))
+    ρ = LinearAlgebra.norm(x)
+    geometric_correction = log(nball_surf_area(ρ, d.n))
     integral_result = -(sqrt(sum((x .- d.c).^2)) - d.r)^2 / (2*d.w^2)
-    result = integral_result + normalization
+    result = integral_result + pdf_normalization - geometric_correction
```
tldrma: I use the formula used in the paper to correct for the geometric scaling, but it seems like it beings to fail at higher dimensions rather quickly.

```
julia> gs2d_integral
1.0345 ± 0.0026

julia> gs3d_integral
1.0694 ± 0.0022

julia> gs4d_integral
1.8762 ± 0.0073
```
where gs2d_integral is the result of bat_integrate with default settings for a 2d gaussian shell density, and so on.

VasylHafych: Why do you think that your integral should be equal to one? As you can see in the last equation (taken from [here](https://arxiv.org/pdf/1808.08051v1.pdf)), the true integral is not one and so you should compare your integration results to that one (given by the last equation): 
Screenshot 2020-07-19 at 13 33 10

VasylHafych:
Hi @tldrma. I have checked your code and integrals of iid samples are correct, so your sampling and integration are working correctly. 

If you would like to have the final integral equal to one, you should add one additional normalization correction. For example, I have tested the following changes: 

```julia
struct GaussianShell{T<:Real, V<:AbstractVector} <: ContinuousMultivariateDistribution
    r::T
    w::T
    c::V
    n::Int
    norm::AbstractFloat # additional normalization that should be computed once in the constructor 
end

function GaussianShell(;r::Real=5, w::Real=2, n::Integer=2)
    c = zeros(n)
    r,w = promote(r, w)

    # normalization constant is given as a 1D integral that needs to be computed once:
    func(x) = x[1]^(n-1)*exp(-(x[1] - r)^2 / (2*w^2))
    norm = log((sqrt(2)*π^((n-1)/2)) / (gamma(n/2)*w)*HCubature.hcubature(func, [0], [r+w*20])[1])

    GaussianShell(r, w, c, n, norm)
end

function Distributions._logpdf(d::GaussianShell, x::AbstractArray)
    pdf_normalization = -log(sqrt(2*π*(d.w)^2))
    integral_result = -(sqrt(sum((x .- d.c).^2)) - d.r)^2 / (2*d.w^2)
    result = integral_result + pdf_normalization - d.norm # subtract additional normalization
    return result
end
```
This works correctly (in a range of 2-7 dimensions, after that `AdaptiveRejectionSampling` cannot generate samples). Here is an example: 

```julia
dist = GaussianShell(r=5, w=2, n=5)
samples_iid = bat_sample(dist, 10^5).result
bat_integrate(samples_iid).result
```
this gives
```terminal
1.0053±0.0034
```
I hope this will be helpful!

oschulz:
> If you would like to have the final integral equal to one

Yes, please include the normalization, `GaussianShell` should be a "real" distribution, it's pdf should have an integral of one.

oschulz:
Thanks! Let me know when it's ready to be merged.

src/distributions/gaussian_shell.jl
```diff
@@ -0,0 +1,71 @@
...
+function rand_nball_surf_samples(rng::AbstractRNG, ndims::Integer, n::Integer)
+    A = randn(rng, ndims, n)
+    A ./= sqrt.(sum(A .* A, dims = 1))
+end
+
+function Distributions.rand(rng::AbstractRNG, d::GaussianShell, n::Int)
+    Distributions._rand!(rng, d, Matrix{eltype(d)}(undef, length(d), n), n)
+end
+
+function Distributions._rand!(rng::AbstractRNG, d::GaussianShell, X::AbstractMatrix, n::Int)
```
oschulz, line 54: This one shouldn't have an argument `n`.

src/distributions/gaussian_shell.jl
```diff
@@ -0,0 +1,71 @@
...
+    X = rand_nball_surf_samples(rng, d.n, n)
+    R = gauss_shell_radial_samples(rng, d.n, d.r, d.w, n)
+    X .* R'
```
oschulz, line 57: `_rand!` is supposed to fill `X` with the result, so that should probably be something like

    X .*= R'
    return X

src/distributions/gaussian_shell.jl
```diff
@@ -0,0 +1,71 @@
...
+    A = randn(rng, ndims, n)
+    A ./= sqrt.(sum(A .* A, dims = 1))
+end
+
+function Distributions.rand(rng::AbstractRNG, d::GaussianShell, n::Int)
+    Distributions._rand!(rng, d, Matrix{eltype(d)}(undef, length(d), n), n)
+end
+
+function Distributions._rand!(rng::AbstractRNG, d::GaussianShell, X::AbstractMatrix, n::Int)
+    X = rand_nball_surf_samples(rng, d.n, n)
```
oschulz, line 55: Write a function `rand_nball_surf_samples!(rng, d.n, X)` that fills `X` instead.

tldrma:
As far as functionality, I think everything works. I'm unsure of a couple things though:

1. I was reading https://docs.julialang.org/en/v1/manual/performance-tips/index.html#man-performance-abstract-container-1 and I'm wondering if what I did it acceptable, namely 
```
struct GaussianShell{T<:Real, V<:AbstractVector, F<:Float64} <: ContinuousMultivariateDistribution
    r::T
    w::T
    c::V
    n::Int
    lognorm::F
end
```
2. Even though random appears to be working, I'm concerned I didn't implement the correct rand functions

```
function Distributions.rand(rng::AbstractRNG, d::GaussianShell, n::Int)
    Distributions._rand!(rng, d, Matrix{eltype(d)}(undef, length(d), n), n)
end

function Distributions._rand!(rng::AbstractRNG, d::GaussianShell, X::AbstractMatrix, n::Int)
    X = rand_nball_surf_samples(rng, d.n, n)
    R = gauss_shell_radial_samples(rng, d.n, d.r, d.w, n)
    X .* R'
end
```

because if I try to change the Int type annotations to the Integer type, then Julia thinks it's too ambiguous.

3. Should I put the empirical `Statistics.cov` function in that you mentioned? You said it won't be necessary for the sampling later, so I haven't make another PR for it yet.

4.  I think the variable names could be a bit better, especially for the normalizations. Maybe something like pdf_normalization and geometric_normalization?

src/distributions/gaussian_shell.jl
oschulz: `n` is implicitly `size(X, 2)`, by definition.

src/distributions/gaussian_shell.jl
```diff
@@ -0,0 +1,71 @@
...
+nball_surf_area(r, ndims) = 2 * π^(ndims/2) / gamma(ndims/2) * r^(ndims-1)
+gs_pdf_r(r, ndims, r0, w) = nball_surf_area(r, ndims) * 1/(2π * w^2) * exp(-(abs(r) - r0)^2 / (2 * w^2))
+
+function gauss_shell_radial_samples(rng::AbstractRNG, ndims::Integer, r0::Real, w::Real, n::Integer)
+    f(r) = gs_pdf_r(r, ndims, r0, w)
+    sampler = AdaptiveRejectionSampling.RejectionSampler(f, (10^-10, Inf))
+    AdaptiveRejectionSampling.run_sampler!(rng, sampler, n)
+end
+
+function rand_nball_surf_samples(rng::AbstractRNG, ndims::Integer, n::Integer)
```
oschulz, line 45: Replace with `rand_nball_surf_samples!(rng::AbstractRNG, X::AbstractMatrix)`

`ndims` will implicitly be `size(X, 1)`.

src/distributions/gaussian_shell.jl
```diff
@@ -0,0 +1,71 @@
...
+    A = randn(rng, ndims, n)
```
oschulz, line 46: Replace with `randn!(rng, X)`

src/distributions/gaussian_shell.jl
```diff
@@ -0,0 +1,71 @@
...
+    A ./= sqrt.(sum(A .* A, dims = 1))
```
oschulz, line 47: Should be

    X ./= sqrt.(sum(X .* X, dims = 1))

then.

src/distributions/gaussian_shell.jl
```diff
@@ -0,0 +1,71 @@
...
+# Arguments
+- `r::Real`: The radius of the Gaussian shell distribution.
+- `w::Real`: Variance of the Gaussian shell distribution.
+- `n::Int`: Number of dimensions.
+
+Constructors:
+```julia
+BAT.GaussianShell(r::Real, w::Real, c::Vector, n::Int)
+"""
+struct GaussianShell{T<:Real, V<:AbstractVector, F<:Float64} <: ContinuousMultivariateDistribution
```
oschulz, line 20: This needs to be `F<:AbstractFloat`. Better use `V<:AbstractVector{T}` or `V<:AbstractVector{<:Real}`.

src/distributions/gaussian_shell.jl
```diff
@@ -0,0 +1,71 @@
...
+end
+
+function Distributions._rand!(rng::AbstractRNG, d::GaussianShell, X::AbstractMatrix, n::Int)
+    X = rand_nball_surf_samples(rng, d.n, n)
+    R = gauss_shell_radial_samples(rng, d.n, d.r, d.w, n)
+    X .* R'
+end
+
+function Distributions._logpdf(d::GaussianShell, x::AbstractArray)
+    pdf_normalization = -log(sqrt(2*π*(d.w)^2))
```
oschulz, line 61: Maybe include this in `lognorm`? And document the calulation of `lognorm` a bit (what each element of the log-sum stands for).

src/distributions/gaussian_shell.jl
```diff
@@ -0,0 +1,71 @@
...
+    integral_result = -(sqrt(sum((x .- d.c).^2)) - d.r)^2 / (2*d.w^2)
```
oschulz, line 62: This deserves a short comment, to explain what's calculated here.

oschulz:
Put the "dummy" cov calculation in for now, I'll take it out later.

oschulz:
Ok, I'm merging this now. @VasylHafych, @salvolc  , after it's merged, can check that it works for you?

PR #151, opened 2020-06-27
Problem with truncated uniform distribution 

VasylHafych:
Uniform distribution does not have `logtp`, so I was not able to truncate it:
```julia
ntdist = NamedTupleDist(a = Uniform(-1,1))
prior = convert(AbstractDensity, ntdist)
bounds = BAT.HyperRectBounds([0..1,], BAT.hard_bounds)
tr_dist = BAT.truncate_density(prior, bounds);
```
```terminal
type Uniform has no field logtp
Stacktrace:
 [1] getproperty(::Uniform{Float64}, ::Symbol) at ./Base.jl:33
 [2] truncate_dist_hard(::Uniform{Float64}, ::Interval{:closed,:closed,Float64}) at /Users/vhafych/MPP-Server/gitrepos/BAT.jl/src/densities/truncated_density.jl:78
```
For the `Uniform` distribution `logrenorm` can be defined as `log(trunc_dist.b - trunc_dist.a) - log(dist.b - dist.a)`. 
Can we merge this workaround?

### DIFF
--- a/src/densities/truncated_density.jl
+++ b/src/densities/truncated_density.jl
@@ -75,7 +75,11 @@ function truncate_dist_hard(dist::Distribution{Univariate}, bounds::Interval)
     hi = clamp(max(lo, maximum(bounds)), min_lo, max_hi)
 
     trunc_dist = truncated(dist, lo, hi)
-    logrenorm = trunc_dist.logtp
+    if dist isa Uniform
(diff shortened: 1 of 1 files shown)


oschulz:
> `if dist isa Uniform ...`

That would be a very Python-like way of doing it. :-)

I tried to fix it in a bit more of a generic fashion, is on master now.

Issue #146, opened 2020-06-24
Mapping between names and indices fails for constant prior parameters 


Cornelius-G:
I just noticed that when setting parameters of the prior constant, the mapping between parameter names and indices for the samples fails (which is needed in the plotting recipes). This is because in the shape of the samples the constant parameter is there, while there are no entries for it in the vector of `unshaped.(samples)`. 
I guess I could fix this by adding some checks to see whether a shape contains a `ConstValueShape` in the function ` asindex(vs::NamedTupleShape, name::Symbol)`in [valueshapes_utils.jl](https://github.com/bat/BAT.jl/blob/master/src/plotting/valueshapes_utils.jl), but I was wondering whether it is a good idea to drop the fixed parameter from the unshaped samples instead of explicitly filling the vectors with the fixed value. I guess it might be needed when using the samples for further computations.

oschulz:
We actually do want to keep constant parameters our of the unshaped samples - AHMI, for example, wouldn't work if it saw constant parameters. The "contract" is that the parameter indices number the degrees of freedom. Also optimizers and similar algorithms don't like to deal with constant values, best if they don't see them. Same for the MCMC samplers.

Computations that use all parameters (also constant ones) should be carried out on the shaped parameters.

For plotting, we can throw an exception for now if const parameters are accessed. Long-term, we could actually access the shaped parameters, if the user specifies a parameter by name instead of by index.

oschulz:
Closed in #153

PR #153, opened 2020-06-29
Fix mapping between indices and names for plotting when some parameters are fixed

Cornelius-G:
This should fix the problem described in issue #146.
(We should still discuss sometime if some of the things in [/src/plotting/valueshapes_utils.jl](https://github.com/Cornelius-G/BAT.jl/blob/f51d092922dc0096d168b892c1ba120e87eba819/src/plotting/valueshapes_utils.jl) can be made a bit nicer and maybe go into [ValueShapes.jl](https://github.com/oschulz/ValueShapes.jl))


oschulz:
> We should still discuss sometime if some of the things in /src/plotting/valueshapes_utils.jl can be made a bit nicer and maybe go into ValueShapes.jl

Yes, a lot of that definitely should, just need to find some time to sort through it. :-)

oschulz:
Thanks!

PR #158, opened 2020-07-05
Adjusted errors used in test_integration.jl to no longer fail in trav…

tldrma:
Tested 300x and no failures.


tldrma:
The updated errors still had a failure in the travis build when testing multimodal cauchy, so I ran it 500 more times to check stability further. Only one error with MvNormal and none with multimodal cauchy, so I just increased all error tolerances by a little bit to give it more padding.

oschulz:
Thanks!

PR #161, opened 2020-07-09
[WIP] Output of summary

Cornelius-G:
This is still pretty much work in progress. 
But I wanted to open this for discussion and suggestions on the general design structure, before I continue coding in a completely wrong direction.

- My idea was to have a new `Summary` type that is returned from `bat_sample` as the second argument: `samples, summary, ... = bat_sample(...)`. There is also another new (abstract) type `SamplerInfo` that is part of `Summary` and is used to store algorithm-specific information.  

- Currently, there is just plain txt output but when we have decided on the general design, I can of course add things like HTML for jupyter notebooks. 

- We should also discuss if there is further information we want to print in this summary by default
- And I guess we need to implement `Base.show()` for some of our types to make, for example, the printing of priors a bit nicer.

short example:
```
using BAT
using Distributions
using IntervalSets

likelihood = params -> begin
    r1 = logpdf.(Normal(-10.0, 1.2), params.a)
    r2 = logpdf.(Normal(-5.0, 0.8), params.c[1])
    r3 = logpdf.(Normal(12.0, 1.4), params.c[2])

    return LogDVal(r1+r2+r3)
end

prior = BAT.NamedTupleDist(
    a = Normal(-3, 4.5),
    b = 0,
    c = [-20.0..20.0, -10..10]
)

posterior = PosteriorDensity(likelihood, prior);

samples, summary, chains = bat_sample(posterior, (10^5, 4), MetropolisHastings());
display(summary)

# write summary to txt-file
io = open("summary.txt", "w")
show(io, "text/plain", summary)
close(io)
```
This would currently result in:
```
BAT.jl - Result Summary
=========================

Model
======
 likelihood:  BAT.GenericDensity{var"#13#14"}(var"#13#14"())
              defined in "~/BAT.jl/examples/dev-internal/output_examples.jl", line 5

 prior:       a = Normal{Float64}(μ=-3.0, σ=4.5)
              b = ValueShapes.ConstValueDist{Univariate,Int64}(value=0)
              c = Product{Continuous,Uniform{Float64},Array{Uniform{Float64},1}}(v=Uniform{Float64}[Uniform{Float64}(a=-20.0, b=20.0), Uniform{Float64}(a=-10.0, b=10.0)])

Sampling
=========================
 algorithm: MetropolisHastings(BAT.MvTDistProposal(1.0), RepetitionWeighting{Int64}())

 total number of samples:      400000
 number of chains:             4
 number of chains tuned:       4
 number of chains converged:   4
 number of samples per chain:  100000

Results of sampling
=========================
fixed: 
--------
  b = 0

1. a:
------
  mean ± std.dev. = -9.5356 ± 1.1589
  global mode     = -9.4647
  marginal mode   = -9.475

2. c[1]:
---------
  mean ± std.dev. = -4.9976 ± 0.79675
  global mode     = -5.0102
  marginal mode   = -4.975

3. c[2]:
---------
  mean ± std.dev. = 9.3727 ± 0.54872
  global mode     = 9.9983
  marginal mode   = 9.975

covariance matrix: 
-------------------
  1.343      -0.00371   -0.0001344
 -0.00371     0.6348    -0.001596 
 -0.0001344  -0.001596   0.3011   
```


oschulz:
We should think about an alternate output style automatically chosen if the number of parameters is large (e.g. no single parameters, no covariance).

oschulz:
Maybe we'll need an higher-level command than `bat_sample` for this - the output we want here goes beyond sampling.

oschulz:
We could add a high-level function `bat_posterior(likelihood, prior)` with a default algorithm `MCMCPosterior(nsamples, ...)` or so. It could return something like `(result = SampledDensity, ...)` or so. `SampledDensity <: AbstractDensity` would wrap the density/posterior definition, the samples, and might additional information like modes. It would also be directly reusable as a prior.

Cornelius-G:
> Maybe we'll need an even higher-level function than `bat_sample` for this - the output we want here goes beyond sampling.

Hm, does it really go beyond? I thought what we wanted to provide is a single command to print out all the information that is of interest for a specific run of the sampling (similar to the [results](https://bat.mpp.mpg.de/pages/tut/0.9/counting_experiment/step2/CountingExp_results.txt) of the C++ BAT).
I don't quite see why we would need an extra high-level function for this that does not even sample 🤔

Cornelius-G:
I guess to proceed here we need to discuss at least the following things:

* which information do we want to show?
I guess the minimal information for sampling should be things like mean, mode, variance/covariance of the individual parameters. But more information about convergence etc. would also be good to have.

But I guess currently main question is:
* how do we solve this technically?
So, which datatype will be used to hold all the information and will be used to implement `Base.show` on?
Currently, I proposed that `bat_sample` returns a `Summary` object which can then be printed.
But I'm not quite sure about what you @oschulz proposed in the comments before.

I was just thinking about it again and thought that it might also be nice to have some mechanism that would allow to add more information to all of our important methods. So that, for example, all the `bat_dosomething` methods would return a corresponding summary object as a part of the namedtuple that is returned. 
So that e.g.
```
samples, sampling_info, chains = bat_sample(....)
integral, integration_info = bat_integrate(...)
```
etc.
This would be a consistent approach and would allow us to extend it with further information in the future without breaking the compatibility. And it would provide the objects that store detailed and algorithm specific information about the particular runs and to be used for printing.

oschulz:
> which information do we want to show

I'd propose we show (via `show`/`display` method(s)`:

* The number of free parameters (total number of DOF), the total number of chains and samples, and the minimum number of eff. samples or so.

* If the number of parameters is low (say <= 20), the mean, median, standard-quantiles and eff. no of samples for each parameter.

* A warning if the samples are not all tagged as tuned and converged, or if the minimum number of eff. samples is less than 50 and less then two times number of free parameters

> Currently, I proposed that bat_sample returns a Summary object which can then be printed.

Yes, would probably be good to store the same kind of information in a `summary` fields of the return value of `bat_sample`. Actually, that would neatly solve the problem of recalculating the information above, and allow for algorithm-specific summary display.

The summary should probably be based on `NamedTuple`s itself, otherwise we may need a custom type for each algorithm, and possibly multiple different types if algorithms can be combined in multiple ways. So we can't specialize `show` for the summary itself, would be hard to avoid type piracy. But we *can*  specialize `show` for whole `NamedTuple` returned by `bat_sample`, since BAT owns the type of the `result = ...::DensitySampleVector` field.

> So that, for example, all the bat_dosomething methods would return a corresponding summary object as a part of the namedtuple that is returned

I like the idea of having a summary field in the `NamedTuple` returned by `bat_...` functions in general. Maybe let's focus on `bat_sample` for the short term, and build on this concept later on. This should probably always be the second field in `bat_...` return values and always be called `summary` or so for consistency.

Cornelius-G:
> The summary should probably be based on `NamedTuple`s itself, otherwise we may need a custom type for each algorithm, and possibly multiple different types if algorithms can be combined in multiple ways. So we can't specialize `show` for the summary itself, would be hard to avoid type piracy. But we _can_ specialize `show` for whole `NamedTuple` returned by `bat_sample`, since BAT owns the type of the `result = ...::DensitySampleVector` field.

Hm, I'm not quite sure I get this point correctly. You suggest the `Summary` to be a `NamedTuple`? But we then can't dispatch `show` correctly on this? Does it then even make sense to return a summary object if we can't use `show` on this object alone? And how is it possible to dispatch `show` on a specific kind of `NamedTuple`?

I get the point that otherwise there might be several custom types needed, but my current solution was to have the `Summary` store the info all algorithms have in common and then have an additional field where algorithm specific infos can be stored (as subtypes of an `AbstractAlgoSpecificInfo` or so, so it would look somewhat like:
```
struct SamplingSummary
    nsamples::Int
    nfreeparams::Int
    posterior::BATPosterior
    algospecific <: AbstractAlgoSpecificInfo
   ...
end
```
Then we can implement a `show` for the different algorithm specific information (if there is some). 

To me this seems a bit cleaner, but maybe I don't quite get your suggestion.

oschulz:
Hm, the dispatch on a `NamedTuple` based on the type of one field could indeed get tricky ... ok, how about this, then:

```julia
struct SamplingSummary{NT<:NamedTuple}
    s::NT
end
```

Keeps it flexible, and we can dispatch on it. We can have one summary type per `bat_...` function.

oschulz:
Maybe the best solution is really to do the proper thing and finally add

```julia
struct SampledDensity{D<:AbstractDensity,S<:DensitySampleVector,...}
    density::D
    samples::S
    effnosmpls::...
    ...
end

Let's do that - initially, we just implement `show` on it, but we'll make it a real, reusable density in the long run.

**Comment by Cornelius-G** [comment]
Ok, yes this sounds good. 
Should `SampledDensity` also have a field with information on how it was sampled? So with `algorithm`, `nchains`, convergence info etc?

**Comment by oschulz** [comment]
I think it could include the chains directly, to allow production of more samples in the future. Convergence info is already part of the samples.

**Comment by oschulz** [comment]
And we add `bat_posterior(likelihood, prior)` that'll call `bat_sample` internally and directly produce a `SampledDensity`.

**Comment by oschulz** [comment]
And `plot(d::SampledDensity)` could also be possibly, simply using the plotting recipes for the samples internally.

**Comment by Cornelius-G** [comment]
> I think it could include the chains directly, to allow production of more samples in the future. Convergence info is already part of the samples.

Ok, but I guess we still need to wrap the chains and so on in a field for sampler-specific information since not all samplers have chains.

**Comment by Cornelius-G** [comment]
> And we add `bat_posterior(likelihood, prior)`, it'll call `bat_sample` internally and directly produce a `SampledDensity`.

Hm ok, but I don't quite see why we would "hide" the sampling inside `bat_posterior`. Wouldn't this make customizing sampler settings etc. complicated? Wouldn't it be easier to make `bat_sample`return a `SampledDensity` (as one of the return values)?

**Comment by oschulz** [comment]
> Ok, but I guess we still need to wrap the chains and so on in a field for sampler-specific information since not all samplers have chains

It would be a `samplesource` or so field with an arbitrary type, so for non-mcmc samplers it wouldn't be a vector of chains. Longer term I want to make that more structured anyhow, to have a supertype for something that can produce non-IID samples.

**Comment by oschulz** [comment]
> Hm ok, but I don't quite see why we would "hide" the sampling inside bat_posterior.

Because it would do more than sampling, it would also calculate statistics, modes, eff. no of samples, etc. - some of which can be done by different algorithms. 

> Wouldn't this make customizing sampler settings etc. complicated?

No, we have to move these into the sampler algorithm types anyhow, things like space partitioning also wrap samplers and the user must be able to pass parameters through. `bat_posterior` would actually take several algorithms as arguments (for sampling, mode-finding, eff-no-of-sample estimation, etc.).

**Comment by Cornelius-G** [comment]
Ah I see. Alright, I will then try to start implementing this with a `SampledDensity`

**Comment by Cornelius-G** [comment]
Ok, so I started implementing `SampledDensity`. It's still WIP but I wanted to check with you @oschulz before proceeding.
```
struct SampledDensity{D<:AbstractDensity,S<:DensitySampleVector, SI<:SamplerInfo}
    density::D
    samples::S
    stats::MCMCBasicStats
    samplerinfo::SI
end
```
There are now several functions defined on `SampledDensity`, like `mean, mode, cov, std, varshape, numberofsamples` etc.
These functions return user-friendly types, so plain numbers or `NamedTuples` (e.g. for `mean, mode, std`) or even `NamedArray` (for `cov` and `cor`): 
```
mean(sd)
(a = -9.528332921881542, c[1] = -5.001565442289139, c[2] = 9.370162159566002)

cor(sd)
3×3 Named Array{Float64,2}
cor ╲  │            a          c[1]          c[2]
───────┼─────────────────────────────────────────
a      │          1.0    0.00650817   -0.00160045
c[1]   │   0.00650817           1.0  -0.000472535
c[2]   │  -0.00160045  -0.000472535           1.0

```
There is also a function to get a `TypedTable` with the parameter estimates (which is also used for printing, see below).

`Base.show` is overloaded and currently produces something like:
```
BAT.jl - SampledDensity
──────────────────────────────

Sampling:
─────────────────────────
algorithm: MetropolisHastings(BAT.MvTDistProposal(1.0), RepetitionWeighting{Int64}())

number of chains:             4
number of chains tuned:       4
number of chains converged:   4
number of samples per chain:  100000

total number of samples:      400000
effective number of samples: (a = 77007.19593461433, c[1] = 76060.6540459323, c[2] = 66417.603476382)


Fixed parameters:
─────────────────────────
number of fixed parameters: 1
fixed parameters: (b = 0,)


Parameter estimates:
─────────────────────────
number of free parameters: 3

Table with 5 columns and 3 rows:
     parameter  mean      std       global_mode  marginal_mode
   ┌──────────────────────────────────────────────────────────
 1 │ a          -9.52642  1.1582    -9.47918     -9.525
 2 │ c[1]       -4.9953   0.801954  -5.01088     -4.975
 3 │ c[2]       9.36853   0.555009  9.99863      9.975


Covariance matrix:
─────────────────────────
3×3 Named Array{Float64,2}
cov ╲  │            a          c[1]          c[2]
───────┼─────────────────────────────────────────
a      │      1.34142   -0.00125221    0.00064486
c[1]   │  -0.00125221      0.643131  -0.000438543
c[2]   │   0.00064486  -0.000438543      0.308035
```

Of course there are still things to extend like overloading plotting for `SampledDensity`, which I will do soon.

File: src/sampleddensity/print_sampleddensity.jl [hunk 1]
```diff
@@ -0,0 +1,62 @@
+function Base.show(io::IO, mime::MIME"text/plain", sd::SampledDensity)
```
oschulz, line 1: But the printing in the same file as the `SampledDensity`, then we don't need a subdir for it.

src/sampleddensity/sampleddensity.jl
```diff
@@ -0,0 +1,126 @@
+export SampledDensity
+
+
+struct SampledDensity{D<:AbstractDensity,S<:DensitySampleVector, SI<:SamplerInfo}
+    density::D
+    samples::S
+    stats::MCMCBasicStats
```
oschulz, line 7: Let's name these `_stats` and `_samplerinfo` to mark them as internal, we want to be able to change the implementation later on.

src/sampleddensity/sampleddensity.jl
```diff
@@ -0,0 +1,126 @@
...
+    samplerinfo::SI
```
oschulz, line 8: Let's use `SI<:Union{Missing,<:AbstractSamplingAlgorithm}` then we don't need the new type `GenericSamplerInfo` at all. Maybe we'll store the chains directly in the future or so, and it's internal anyway.

src/sampleddensity/sampleddensity.jl
```diff
@@ -0,0 +1,126 @@
...
+    modes = sd.stats.mode
+    return (; zip(active_keys(sd), modes)...,)
+end
+
+function marginalmode(sd::SampledDensity)
+    modes = unshaped(bat_marginalmode(sd.samples).result)
+    return (; zip(active_keys(sd), modes)...,)
+end
+
+import Statistics.cov
```
oschulz, line 88: Just write `function Statistics.cov(...`, no need for `import Statistics.cov`

Same for the other specializations below.

src/samplers/importance/importance_sampler.jl
```diff
@@ -45,11 +49,12 @@ function bat_sample_impl(
 
     samples = _gen_samples(algorithm, n, truncated_density)
 
-    logvals = logvalof.(Ref(truncated_density), samples)
+    logvals = logvalof_unchecked.(Ref(truncated_density), samples)
```
oschulz, line 52: Samplers should always call `logvalof`, not `logvalof_unchecked`. In any case, this should go in a separate PR. :-)

src/samplers/mcmc/mcmc_algorithm.jl
```diff
@@ -90,6 +90,10 @@ end
     converged::Bool
 end
 
+struct MCMCSamplerInfo <: SamplerInfo
```
oschulz, line 93: This type shouldn't be necessary, since the algorithms can be recovered from the chains.

src/samplers/importance/importance_sampler.jl
Cornelius-G: Oops. Did not change this on purpose. Might be a remnant from an older commit that was not properly updated by git. Will change this back.

src/sampleddensity/sampleddensity.jl
Cornelius-G: Hm, but we need to somehow store the chains in the `SampledDensity`. They are of type `MCMCIterator`.
But we only need to do this if it was a MCMC sampler. Otherwise we might want to store some other infos about the non-MCMC samplers. That's why I introduced this `SamplerInfo` type. For MCMC algos this basically just wraps the chains. Just storing the `AbstractSamplingAlgorithm` would not be sufficient for this, or would it?

src/samplers/mcmc/mcmc_algorithm.jl
Cornelius-G: Hm, yes. But see my comment above.

src/sampleddensity/sampleddensity.jl
oschulz: Ah - well, longer term I had planned to add a `AbstractSampleGenerator` type, subtypes of which would store chains, etc., and allow for production of more samples as well. So maybe just add that now, with a subtype `MCMCSampleGenerator` (storing the vector of `MCMCIterator`s), without any special functionality get. It should support a function `getalgorithm` or so.

src/samplers/mcmc/mcmc_algorithm.jl
oschulz: Ok, see above.

src/sampleddensity/sampleddensity.jl
oschulz: `SampledDensity.generator` can then become a user-accessible field, and `AbstractSampleGenerator` and `MCMCSampleGenerator` can be documented and exported. However, `MCMCSampleGenerator._chains` should be an internal field, not part of the stable API.

Cornelius-G: Ok, I tried to introduce the `AbstractSampleGenerator` type.
For MCMC it just wraps the chains. For all other samplers it currently just contains the algorithm type.

src/sampleddensity/print_sampleddensity.jl
Cornelius-G: I put everything inside one file now. But I'm not sure where we should put this file, so it is currently still in its own subdir. 
I tried to put it into the densities folder but there we have the problem that things like the `MCMCStats` and so on are not yet defined, so one would have to forward-declare or so. What do you propose?

oschulz: > But I'm not sure where we should put this file

I'll become a fully featured density later, so it goes in "densities", of course. :-)

Cornelius-G: Ok, I moved this now. But had to include the file directly in `src/BAT.jl` because there were problems with missing declarations of types like `MCMCStats` and the sampling algorithm types. Is there a better way to do this?

oschulz: Ah, darn, right, the algo defs come after. Ok, then I'd say let's move it into the "samplers" directory for now.

Cornelius-G:
Hm, I just pushed a commit moving the file `sampled_density.jl` to `src/samplers`. On my fork this was updated but the PR somehow does not show this change.

oschulz:
> On my fork this was updated but the PR somehow does not show this change.

Yep, I see it on your fork. Just delete this PR and make a new one.

PR #163, opened 2020-07-19
Kolmogorov-Smirnov Goodness of Fit Test

VasylHafych:
Hi @oschulz, @salvolc. I have improved `bat_compare` function so that it can now use different ways to treat a number of samples is the Kolmogorov-Smirnov test. By default, it will use an effective number of samples (alternatively, the length of the sample vector can be used, or the sum of its weights). This function can be used now to run final benchmarks for the paper. 

@salvolc, updated `bat_compare` function works in the same way as the previous one. The only small difference is that results are returned in a form of a `NamedTuple`, so you should run `bat_compare(sample_1, sample_2).result`.

### DIFF
--- a/docs/src/experimental_api.md
+++ b/docs/src/experimental_api.md
@@ -6,6 +6,7 @@ Instead, compatibility is only guaranteed across changes in patch version, but
 *not* across changes of minor (or major) version changes.
 
 ```@docs
+bat_compare
 BAT.ExternalDensity
 BAT.FunnelDistribution
 BAT.MultimodalCauchy
--- a/src/statistics/ks_test.jl
+++ b/src/statistics/ks_test.jl
@@ -1,14 +1,5 @@
 # This file is a part of BAT.jl, licensed under the MIT License (MIT).
 
-function ApproximateTwoSampleKSTest(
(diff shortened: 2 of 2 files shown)


oschulz:
> (alternatively, the length of the sample vector can be used, or the sum of its weights)

Let's take those options out, the eff. number of samples should always be the (only) correct choice.

VasylHafych:
If you have a lot of samples in many dimensions (~10^7) then the estimation of the effective number of samples can be quite long. I decided to keep these options to be able to have a fast check. Using the length of the vectors just makes KS criteria more strict, so if samples are the same it still gives reasonable results.

oschulz:
Ok, we can merge it like this, we should replace that by a choice of eff-sample-site estimators, one of which can then be a quick and inaccurate one.

Can this be merged?

VasylHafych:
> Ok, we can merge it like this, we should replace that by a choice of eff-sample-site estimators, one of which can then be a quick and inaccurate one.

Sure, we can implement this later. 

> Can this be merged?

Yes!

oschulz:
Thanks!

PR #165, opened 2020-07-25
Plotting now allows to specify bin edges

Cornelius-G:
Binning of plots can now not only be passed as the number of bins but also by specifying the bin edges as a `Range`.
Examples:
```
plot(samples, :a, bins = -10:0.1:10)
plot(samples, (:a, :b), bins = (50, -10:0.1:10))

plot(samples, bins =  (50, -10:0.1:10))
plot(samples, bins = (a = 50, b = -5:0.5:5))
```
Also,  `vsel` now accepts indices and parameter names.

@oschulz The only problem we now have is that the syntax in the fourth line of the example code above does not work for multivariate parameters, since expressions like e.g. `:(b[1])` are not allowed as names in a `NamedTuple`. 
I remember we already discussed going back to something like `b_1` etc. So maybe we should consider doing this now?

### DIFF
--- a/docs/src/plotting.md
+++ b/docs/src/plotting.md
@@ -32,7 +32,8 @@ Keyword arguments:
     * `:central_intervals`
     * `:histogram`, alias `:steppost`
     * `:stephist` (default for prior)
-* `nbins::Integer = 200`: number of histogram bins
+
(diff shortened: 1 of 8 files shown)


oschulz:
Thanks!!

PR #168, opened 2020-08-01
[WIP] Include gaussian shell

sthayashi:


oschulz:
Thanks! Can you deal with the test failures?

sthayashi:
I think the missing dependency for `SpecialFunctions.jl` is only used for its `gamma` function in gaussian_shell.jl. Can I add this into the `Project.toml`?

oschulz:
Of course, SpecialFunctions is already an indirect dependency anyway. Put it in with a version bound of "0.10".

oschulz:
Almost there - just listing `BAT.GaussianShell` in the docs is still missing. You can add it to "experimental_api.md" for now.

Hint: You'll have a faster turn-around time if you run tests and docs-gen (see "docs/make.jl") locally before pushing and having Travis run it.

PR #170, opened 2020-08-05
[WIP] Introduce SampledDensity

Cornelius-G:


Cornelius-G:
@oschulz maybe you can take a look and say what needs to be done before merging this.
I guess we can add further functionality also later on.

oschulz:
Thanks for this nice work!

You're right, we can always add more stuff - for now we can merge it and start playing with it. :-)

PR #180, opened 2020-08-08
Multimodal cauchy assertion

sthayashi:


src/distributions/multimodal_cauchy.jl
```diff
@@ -25,6 +25,9 @@ struct MultimodalCauchy{M<:MixtureModel, P<:Product}  <: ContinuousMultivariateD
 end
 
 function MultimodalCauchy(;μ::Real=1, σ::Float64=0.2, n::Integer=4)
+    if n < 2
+        throw("Must specify dimension greater than 1")
```
oschulz, line 29: Use

    @argcheck n < 2

or

    throw(ArgumentError("Minimum number of dimensions for MultimodalCauchy is 2")

and add a test

    @test_throws ArgumentError MultimodalCauchy(n = 1)

PR #184, opened 2020-08-14
Density sample tests

sthayashi:


sthayashi:
Question regarding intended functionality of `DensitySample` and `DensitySampleVector`.

Should `mean`, `std`, and `median` of a `DensitySampleVector` all return the typical weighted version? When I code it myself, I get the same weighted mean, but I get a different weighted standard deviation.

```
dsv = DensitySampleVector(...)
X = flatview(dsv.v)
w = dsv.weight
weighted_mean = sum(X .* w' , dims=2)/sum(w)
weighted_std = sqrt.(sum((X .- weighted_mean).^2 .* w', dims=2) / sum(w))

julia> weighted_mean
5×1 Array{Float64,2}:
 -0.3209832266641278
  0.0033103122539178374
 -0.1966013889679025
  0.465377117235548
 -0.0008382504642858368

julia> mean(dsv)
5-element Array{Float64,1}:
 -0.3209832266641279
  0.0033103122539178374
 -0.19660138896790252
  0.46537711723554803
 -0.0008382504642858368

```
Maybe I'm doing something wrong with the std calculation, but `weighted_mean` and `mean(dsv)` are the same to something like 16 decimal places, which is why I'm kind of suspicious of how `Statistics.std(samples::DensitySampleVector)` is implemented.

oschulz:
> all return the typical weighted version

Yes, they should! Please feel free to investigate.

sthayashi:
I'm not understanding how broadcast should behave for `DensitySample`. As it stands now, it seems like it doesn't actually do anything useful. E.g. 
```
julia> ds = DensitySample([1.0, 2.0, 3.0], 4.0, 5.0, nothing, nothing)
DensitySample{Array{Float64,1},Float64,Float64,Nothing,Nothing}([1.0, 2.0, 3.0], 4.0, 5.0, nothing, nothing)

julia> typeof.(ds)
┌ Info: Broadcast.broadcastable
└   shape = DensitySample{Array{Float64,1},Float64,Float64,Nothing,Nothing}([1.0, 2.0, 3.0], 4.0, 5.0, nothing, nothing)
DensitySample{Array{Float64,1},Float64,Float64,Nothing,Nothing}

julia> typeof( typeof.(ds) )
┌ Info: Broadcast.broadcastable
└   shape = DensitySample{Array{Float64,1},Float64,Float64,Nothing,Nothing}([1.0, 2.0, 3.0], 4.0, 5.0, nothing, nothing)
DataType

```
Which seems kind of useless to me. Is broadcasting being used indirectly somewhere else?

oschulz:
> I'm not understanding how broadcast should behave for DensitySample

`DensitySample` is supposed to behave like a scalar under broadcasting. So the expected behavior is

```julia
f.(ds) == f.(Ref(ds)) == f(ds)

f.(ds, A) == f.(Ref(ds), A)
```

oschulz:
It's sufficient to test ´

```julia
Base.Broadcast.broadcastable(ds) == Ref(ds)
```

though.

sthayashi:
I think that this won't work. See: https://github.com/JuliaLang/julia/issues/31813
Since it's just a wrapper, would something like `Base.Broadcast.broadcastable(ds).x == Ref(ds).x` be good enough?

oschulz:
Good catch! Then test

```
@inferred(Base.Broadcast.broadcastable(ds)) isa Ref && Base.Broadcast.broadcastable(ds).x == Ref(ds).x
```

or so.

sthayashi:
Should the line `if minimum(W) / maximum(W) > 10^-2` be `if minimum(W) / maximum(W) > fraction`?
```
function drop_low_weight_samples(samples::DensitySampleVector, fraction::Real = 10^-5)
    W = float(samples.weight)
    if minimum(W) / maximum(W) > 10^-2
        samples
    else
        W_s = sort(W)
        Q = cumsum(W_s)
        Q ./= maximum(Q)
        @assert last(Q) ≈ 1
        ind = searchsortedlast(Q, fraction)
        if ind !== 0
            thresh = W_s[searchsortedlast(Q, fraction)]
            idxs = findall(x -> x >= thresh, samples.weight)
            samples[idxs]
        else
            samples
        end
    end
end
```

sthayashi:
This is line 303 in density_sample.jl

sthayashi:
I ***think*** if we make the previously mentioned change, then we could just get rid of this nested `if`/`else` block, but maybe there's some obvious condition I'm not accounting for.
Seems like this should do the exact same as is currently implemented, but with less code?
```
function drop_low_weight_samples(samples::DensitySampleVector, fraction::Real = 10^-5)
    W = float(samples.weight)
    if minimum(W) / maximum(W) > fraction
        samples
    else
        W_s = sort(W)           
        Q = cumsum(W_s)
        Q ./= maximum(Q)                        
        @assert last(Q) ≈ 1
        ind = searchsortedlast(Q, fraction)
        thresh = W_s[ind]
        idxs = findall(x -> x >= thresh, samples.weight)
        samples[idxs]
    end                                                                               
end
```

oschulz:
> Should the line `if minimum(W) / maximum(W) > 10^-2` be `if minimum(W) / maximum(W) > fraction`?

No, but you make a good point. We should add a separate keyword argument `threshold` or so.

oschulz:
> I _**think**_ if we make the previously mentioned change, then

See above, the 10^-2 and `fraction` play a different role, `fraction` is a probability mass fraction, the other triggers if we do need to drop samples at all.

sthayashi:
I think it's just that samples don't always drop when I would think they should. The test I have implemented now is triggering all parts of that function, but if you look at the three tests written to test this
```
        @test BAT.drop_low_weight_samples(DensitySampleVector(dsv_merged.v, dsv_merged.logd, weight=w1)).v == dsv_merged.v
        @test BAT.drop_low_weight_samples(DensitySampleVector(dsv_merged.v, dsv_merged.logd, weight=w2), minimum(w2)/maximum(w2)).v == dsv_merged.v
        @test BAT.drop_low_weight_samples(DensitySampleVector(dsv_merged.v, dsv_merged.logd, weight=w2), 10^-10).v == dsv_merged.v
```
no samples are dropped in any of the cases.

oschulz:
> I think it's just that samples don't always drop when I would think they should

Maybe do this: Generate some random samples, and set `threshold = 0.1, fraction = 0.05` and `threshold = 10^-6, fraction = 0.05` and `threshold = 0.1, fraction = 10^-6`, and check that the length of the resulting sample vector is only reduced in the first case and that the samples are identical to the input in the other two cases. And that the var of the samples is still (mostly) correct in all cases.

src/parameters/density_sample.jl
```diff
@@ -176,7 +176,9 @@ function DensitySampleVector(
     aux::AbstractVector = fill(nothing, length(eachindex(v)))
 )
     if weight == :RepetitionWeight
-        v, weight = repetition_to_weights(v)
+        v, logval, weight = repetition_to_weights(v, logval)
```
sthayashi, line 179: When the current version of this function is used, it changes dimension of `v` and `weight`, but not the other three properties in a `DensitySampleVector`. I just made this commit to give an example of how it could be fixed by using `logval` as a demonstration. 

Is there a downside to just passing all args into `repetition_to_weights` and repeating what I did for `logval` for `info` and `aux`?

sthayashi: If I understood correctly, something like
```
julia> v
6-element ArrayOfSimilarArrays{Float64,1,1,2,ElasticArray{Float64,2,1,Array{Float64,1}}}:
 [1.0, -0.1, 0.5]
 [-2.5, 0.2, 2.7]
 [-2.5, 0.2, 2.7]
 [1.0, 1.0, 1.0]
 [-2.5, 0.2, 2.7]
 [1.0, 1.0, 1.0]
```
should return
```
julia> (idxs, counts)
([1, 2, 4, 5, 6], [1, 2, 1, 1, 1])
```
and then `repetition_to_weights` should take in a `DensitySampleVector` and returns `idxs` with weights replaced by `counts`?

oschulz: Hm, so far, we've only summed up equal entries that follow each other directly. I think we should keep that behavior. Speaking of which, `info` and `aux` must also be the same for entries that are to be summed up.

sthayashi: I see what you mean about not copying `v`. I think 0f1230aaf4fa2e028ec89615b10b0da54bd7706f does what you asked, but this avoids using a helper function which might mean I misunderstood something.
Not sure if the ctor is stylistically bad though. 
```
function DensitySampleVector(
    v::AbstractVector,
    logval::AbstractVector{<:Real};
    weight::Union{AbstractVector{<:Real}, Symbol} = fill(1, length(eachindex(v))),
    info::AbstractVector = fill(nothing, length(eachindex(v))),
    aux::AbstractVector = fill(nothing, length(eachindex(v)))
)  
    if weight == :RepetitionWeight
        idxs, weight = repetition_to_weights(v)
        return DensitySampleVector((ArrayOfSimilarArrays(v[idxs]), logval[idxs], weight, info[idxs], aux[idxs]))
    else
        return DensitySampleVector((ArrayOfSimilarArrays(v), logval, weight, info, aux))
    end
end
```

oschulz: Well, it does use a helper function, namely `repetition_to_weights`, and that one doesn't handle `info`, etc.

oschulz:
What's the status on this?

oschulz:
Is this still WIP?

sthayashi:
It should be ready to review

oschulz:
Ok, let's see if it still builds, I just updated it with the current master.

PR #191, opened 2020-09-20
Fix AdvancedHMC warmup with StanHMCAdaptor

cescalara:
Some changes to fully implement Stan-style adaptation of HMC params
* `MCMCBurninStrategy` for `AHMC` types with a single, fixed-length cycle
* Use `StanHMCAdaptor` to specify the number of adaptations, `n_adapts`
* Correctly initialize `AdvancedHMC.StanHMCAdaptor` and pass `i`, `n_adapts` in `ahmc_step!()`
* Check that adaptation is complete as well as for convergence in `mcmc_tune_burnin!()`

Still more to improve, but with these changes, the stepsize now gives sensible values compared to Stan and the published AdvancedHMC.jl as shown below.

Screenshot 2020-09-20 at 18 48 26


oschulz:
Oh very nice, thanks a lot!

PR #194, opened 2020-10-01
 Add new AHMCSampleID to store Hamiltonian transition info

cescalara:
Extended `MCMCSampleId` to store extra info from `AdvancedHMC.transition` for each iteration. New fields include `hamiltonian_energy`, `tree_depth` and `divergent`. Also added `step_size`, as while tuned per chain, the `JitteredLeapfrogIntegrator` option leads to independent step sizes for each sample.


oschulz:
Thanks, Francesca!

PR #203, opened 2020-10-09
Prior Plot Bugfix

VasylHafych:
Added flattening of samples. This should give us a prior plot: 
```julia
plot(prior, (1, 3))
```
@Cornelius-G, I do not know if this is intended, but why we dropped `vsel` keyword?


oschulz:
> @Cornelius-G, I do not know if this is intended, but why we dropped vsel keyword?

Uh, not sure, this should definitely use vsel!

oschulz:
Thanks for the bugfix!

PR #205, opened 2020-10-16
tests for algodefaults

sthayashi:


sthayashi:
Would it be all right to write tests for the default algorithms with their already existing tests? E.g. test the default integration with `bat_integrate(some_distribution)` inside of `test_ahmi_integration.jl`? Or should there be a separate `test_algodefaults.jl`?

oschulz:
Probably fine to just do it via the existing test.

sthayashi:
Maybe this should be posted as an issue, but I just while testing the default sampling algorithm, I noticed that the default `bat_sample` for `AbstractDensity` is missing definition of `max_nsamples`. 

https://github.com/bat/BAT.jl/blob/7924f352d2d26dd2d250cdd46df2d584d69e9f42/src/algodefaults/default_sampling_algorithm.jl#L11-L20

Should this be fixed within this PR, or a separate one?

oschulz:
> I noticed that the default bat_sample for AbstractDensity is missing definition of max_nsamples

Ah, yes ... make the number of samples part of the args of `bat_default(::typeof(bat_sample), ...`, so let's make it

```julia
 bat_default(::typeof(bat_sample), ::Val{:algorithm}, ::AbstractDensity, n::Integer)
```

(While you're at it, could you replace any remaining `BAT.AnyNSamples` in BAT by `Integer`?).

Then we can use

```julia
     burnin = MCMCMultiCycleBurnin( 
         max_nsamples_per_cycle = max(div(n, 10), 100), 
         max_nsteps_per_cycle = max_nsamples_per_cycle * 10,
     ), 
```

Oh, and the docs for `bat_sample` need to be updated as well ... forgot about that.

sthayashi:
I just realized that there's also an undefined `max_nsteps` on line 16 in the above code snippet. I think we would want to pass that into the functions here?

https://github.com/bat/BAT.jl/blob/7924f352d2d26dd2d250cdd46df2d584d69e9f42/src/algotypes/sampling_algorithm.jl#L102-L118

Or it should be something calculated from within `bat_default` itself?

oschulz:
Which `max_nsteps`?

sthayashi:
> Maybe this should be posted as an issue, but I just while testing the default sampling algorithm, I noticed that the default `bat_sample` for `AbstractDensity` is missing definition of `max_nsamples`.
> 
> https://github.com/bat/BAT.jl/blob/7924f352d2d26dd2d250cdd46df2d584d69e9f42/src/algodefaults/default_sampling_algorithm.jl#L11-L20
> 
> Should this be fixed within this PR, or a separate one?

Inside of the same function `default_sampling_algorithm.jl` there is no `max_nsamples` nor `maxnsteps` defined.

sthayashi:
Tasks before ready for review
- [ ] Test defaults for transforms

- [ ] Clean commit history

oschulz:
Thanks!

PR #220, opened 2020-12-08
multithread cuba integration test

sthayashi:


oschulz:
Shouldn't be necessary, since we already test with `nthreads = nthreads()`.

PR #229, opened 2020-12-13
GaussianShell sampling fixes

sthayashi:


sthayashi:
Two bugs to fix:

- [ ]  Specify `search_range` of `RejectionSampler`, or intelligently choose an initial starting point. 

- [ ] Fix problem with line segments not being sorted.

oschulz:
@VasylHafych , does this solve the problems you had?

sthayashi:
I've made a few changes which seems to have improved the parameters I've been testing. I've gotten a few sporadic failures, but rerunning the exact same code made it work (will look into more deeply, but at least this seems to solve almost all of the problems for me). If there are still problems, please let me know the parameters you're using on the gaussian shell constructor and I will look into it.

VasylHafych:
Hi @sthayashi. I have just tried this with IID and HMC (up to 100 dimensions) and it seems to work well. Thanks a lot for fixing this.

VasylHafych (approved):

PR #235, opened 2020-12-17
Dirichlet transform

jknollm:


oschulz:
Ok, tests passing now - Ok to merge as it is, @jknollm, or would you still like to add/change stuff?

jknollm:
I think it is fine as it is. It should work and we have a test for it.

oschulz:
Done. :-)

PR #247, opened 2021-01-11
transform tests

sthayashi:


sthayashi:
https://github.com/bat/BAT.jl/blob/d0f48a2436b9ae8217e731648e52b4741004e4ff/src/algotypes/transform_algorithm.jl#L189-L193

Are the arg types in this function correct?

oschulz:
> Are the arg types in this function correct?

Should be, yes.

sthayashi:
I'm unsure how this function should behave, because `getprior` is only defined for a `PosteriorDensity`, and the function to apply a `FullDensityTransform` for a `PosteriorDensity` is already be written for this above. Here are the two relevant functions
https://github.com/bat/BAT.jl/blob/d0f48a2436b9ae8217e731648e52b4741004e4ff/src/algotypes/transform_algorithm.jl#L182-L193

oschulz:
Oh, yes, that's wrong - for a `DistributionDensity`, we want to use the density itself as the "prior" (more like a prior equivalent).

Should be

    trafo = _distribution_density_trafo(target, density)

like in `bat_transform_impl(target::Union{PriorToUniform,PriorToGaussian}, density::DistributionDensity, algorithm::FullDensityTransform)`.

PR #253, opened 2021-02-05
Allow keywords when running UltraNest ?

Cornelius-G:
Or is there a reason why these keywords were disabled?
Also, I'm not sure about `viz_callback` in lines 65 and 139


oschulz:
> Or is there a reason why these keywords were disabled?

Oooops, thanks! Must have happened while debugging of the initial implementation ...

src/samplers/nested_sampling/ultranest.jl
```diff
@@ -122,33 +122,33 @@ function bat_sample_impl(
     end
 
     ndims = totalndof(vs)
-    paramnames = ["v_$i" for i in 1:ndims]
+    paramnames = all_active_names(varshape(density_notrafo))
```
oschulz, line 125: UltraNest only sees the transformed parameter space, so for this to work, `all_active_names` needs to support things like `BAT.all_active_names(ArrayShape{Real}(5))` too. If `all_active_names` gets support for `ScalarShape` and `ArrayShape` (can be done in this PR), that would work.

However, "valueshapes_utils.jl" will need to move from "plots" to "utils", to have a sane order of src includes.

oschulz (changes requested):

src/samplers/nested_sampling/ultranest.jl
Cornelius-G: But  `ScalarShape` and `ArrayShape` don't contain any information on the names, right? 
`all_active_names` works by just accessing the keys of a `NamedTupleShape` (usually coming from the posterior or prior).
For my tests, the code above worked, but I guess because I'm testing on a posterior where the prior has names assigned to the parameters. But the CI tests use a prior that has no names. 
So maybe we can just do something like 
```julia
if isa(varshape(density_notrafo), NamedTupleDist)
paramnames = all_active_names(varshape(density_notrafo))
end
```
?

oschulz: > But ScalarShape and ArrayShape don't contain any information on the names, right?

oschulz: > But ScalarShape and ArrayShape don't contain any information on the names, right?

Yes, but it can just return `"v"` for `ScalarShape` and something like `["v_$i" for i in 1:ndims]` for `ArrayShape`.

Cornelius-G: Yes sure, with generic names that is possible. 
Ah, I see why I had "valueshapes_utils.jl" in "plotting" and not in "utils". It needs to know some of the BAT types.
Any suggestions where to place it/how to include it?

oschulz: Let's split it - the BAT-independent part can go in "utils/", in preparation for moving it into ValuesShapes eventually. The BAT-dependent part can stay in plots or maybe even move closer to the types it depends on.

examples/benchmarks/benchmark_results/results/results_2D.tex
```diff
@@ -0,0 +1,20 @@
+\begin{table}
```
oschulz, line 1: This should not be committed.

examples/benchmarks/benchmark_results/results/results_2D.txt
```diff
@@ -0,0 +1,18 @@
+┌─────────────────┬──────────────────┬────────────────────┬──────────────────┐
```
oschulz, line 1: This should not be committed.

examples/benchmarks/benchmark_results/results/run_stats_2D.tex
```diff
@@ -0,0 +1,9 @@
+\begin{table}
```
oschulz, line 1: This should not be committed.

examples/benchmarks/benchmark_results/results/run_stats_2D.txt
```diff
@@ -0,0 +1,7 @@
+┌──────────────┬────────┬─────────┬───────┐
```
oschulz, line 1: This should not be committed.

oschulz (changes requested):

examples/benchmarks/benchmark_results/results/run_stats_2D.txt
Cornelius-G: Oops, right. Didn't notice these files.

oschulz: Kinda suspected they snuck in there. :-)

oschulz:
Thanks!

PR #262, opened 2021-02-20
Test Cuba integration for failures and multithreading

sthayashi:


test/integration/test_cuba_integration.jl
```diff
@@ -41,5 +41,7 @@ using Distributions, ValueShapes
     test_integration(mvprior_simple, DivonneIntegration(trafo = NoDensityTransform(), rtol = 1e-3))
 
     test_integration(mvprior, CuhreIntegration(trafo = NoDensityTransform(), rtol = 1e-3))
+    
     @test_throws ErrorException test_integration(mvprior, CuhreIntegration(maxevals=4))
+    test_integration(mvprior, CuhreIntegration(maxevals=4, strict=false))
```
sthayashi, line 46: Possible bug on line https://github.com/bat/BAT.jl/blob/224be1e744f70236b62e63b54de1c6994a529acc/src/integration/cuba_integration.jl#L277
I would expect this test to throw an error, but it doesn't.

oschulz: No, with `strict=false` it should just show a warning, not throw an error (but it should throw with `strict = true`).

sthayashi:
Ready for review

oschulz:
Thanks!

PR #265, opened 2021-03-05
Adds test for RandResampling and OrderedResampling test to test_bat_sample.jl

jlrestrepol:
Hi @oschulz (I could not assign you as a reviewer so I hope this is ok)

This PR includes test for the three functions in the bat_samples module and only modifies the test_bat_samples.jl file. 

In this script I added tests for RandResampling and moved the one that existed for OrderedResampling to another test set together with more tests (do you agree with this organisation change?). 

There's one strange behaviour in OrderedResampling that I'm not sure is intended, this behaviour will make the following test fail (that's why it's commented out in line 52 of the script):

`@test length(@inferred(bat_sample(result, OrderedResampling(nsamples = 10^3))).result) == 10^3`

Basically the output vector doesn't have in general the number of samples that's specified with the `n_samples` parameter.  If `nsamples = length(result)` then the test passes.  

**To reproduce the issue:**

- `dist = Normal()`
- `result = (bat_sample(dist, IIDSampling(nsamples = 12))).result`
- `length((bat_sample(result, OrderedResampling(nsamples = 15))).result.v)` Run this several times and you'll get different results

The different lengths come in the source code from only accepting the sample with probability n/m (n = No. samples in result, m No. samples in `OrderedResampling`) (line 126 [src](https://github.com/bat/BAT.jl/blob/eaa6fe8a4fdda48a44d20d932cde8377ca1310b2/src/samplers/bat_sample.jl))

Is this intended?

### DIFF
--- a/test/samplers/test_bat_sample.jl
+++ b/test/samplers/test_bat_sample.jl
@@ -49,7 +49,7 @@ using Random, Distributions, StatsBase
         dist = MvNormal([0.4, 0.6], [2.0 1.2; 1.2 3.0])
         result = @inferred(bat_sample(dist, IIDSampling(nsamples = 10^5))).result
 
-        #@test length(@inferred(bat_sample(result, OrderedResampling(nsamples = 10^3))).result) == 10^3#Commented out: fails but looks intended behav?
+        @test isapprox(mean([length(@inferred(bat_sample(result, OrderedResampling(nsamples = 10))).result.v) for i in 1:10^3]), 10, rtol = 10^-1)
 
         @test @inferred(bat_sample(Random.GLOBAL_RNG, result)).result isa DensitySampleVector#Check that types are consistent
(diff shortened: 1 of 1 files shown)


oschulz:
Thanks! Let me know when it's ready to be merged from your side.

jlrestrepol:
I was just now writing a comment(which I soon publish) explaining what we discussed for future reference. I just pushed a commit that implements what we discussed and from my side this PR is ready to merge.

jlrestrepol:
**Answer to previous question:** The algorithm used for OrderedSampling only returns an array of size equal to `nsamples` when n = m, but the variations of the length should be small and should follow a Poisson distributions. This is a price to be paid for gaining performance. I changed the test to sample many times with `nsamples = 10` and then took the average length of the arrays I got, this should be equal to 10.

oschulz:
Thanks!

PR #268, opened 2021-03-13
Add smoothing of contour plots using Gaussian filtering

Cornelius-G:
Allows smoothing of contour plots using Gaussian filtering (via ImageFiltering.jl).

Without smoothing:
`p = plot(samples, (:a, :c), st=:smallest_intervals_contour, bins=100, smoothing=0)`
no_smoothing

With smoothing:
`p = plot(samples, (:a, :c), st=:smallest_intervals_contour, bins=100, smoothing=1.5)`
with_smoothing

Also possible to smooth the dimensions separately:
`p = plot(samples, (:a, :c), st=:smallest_intervals_contour, bins=100, smoothing=(0, 1.5))`
with_smoothing_y

(Also replaced all remaining "local modes" by "marginal modes")


oschulz:
This is very nice - only, ImageFiltering would be quite a heavy additional dependency for BAT, it pull in several other imaging-related packages that we don't need at all. Can we find a similar functionality/algorithms in a more statistics-oriented Julia package?

oschulz:
We already depend on KernelDensity.jl, could we maybe do this via BivariateKDE?

Cornelius-G:
> We already depend on KernelDensity.jl, could we maybe do this via BivariateKDE?

This was actually also my first idea. However, the things I tried with KDEs (and also with grid interpolations) did not really improve the contours.
I'll see if I can find some more light-weight filtering algorithms.

Cornelius-G:
So, I removed the dependency on `ImageFiltering.jl`  and added a file ` convolution_utils.jl` that contains very basic code for a convolution (including padding) and Gaussian kernels.
The implementation is *really simplistic* and quickly put together from the code [here](https://towardsdatascience.com/understanding-convolution-by-implementing-in-julia-3ed744e2e933) and [here](https://github.com/JuliaImages/ImageFiltering.jl/blob/master/src/kernelfactors.jl#L411), but seems to work.
Unfortunately, I currently don't have the time to make this much nicer, but maybe someone else would like to do a bit of code reduction and profiling.

However, there are efforts for separating the filtering stuff from the image-related code in `ImageFiltering.jl`  being currently [discussed](https://github.com/JuliaImages/ImageMorphology.jl/issues/40#issuecomment-778045150). So, we might want to keep an eye on that and switch back to a more sophisticated implementation later.

Cornelius-G:
@oschulz GitHub Actions had some issues yesterday evening, so the CI failed or is still queued. I don't see the button to restart CI. Can you run this again somehow?

oschulz:
Yes, I was wondering ... restarting it now.

oschulz:
Nice!

PR #271, opened 2021-03-30
Fix distribution transforms for cases with non-trivial effective NDOF

oschulz:


oschulz:
CC @cescalara

cescalara:
Thanks! Will check this out

cescalara:
Works nicely! For reference here is a minimal example with a Dirichlet prior.

```Julia
using BAT, Distributions, ValueShapes

# Generate some fake data from a Gaussian mixture
N = 1000
theta = rand(Dirichlet([3, 1]))
x = Array{Float64}(undef, N)
for i in 1:N
    comp = rand(Categorical(theta))
    if comp == 1
        x[i] = rand(Normal(2, 1))
    end
    if comp == 2
        x[i] = rand(Normal(5, 1))
    end
end

# Data to pass to BAT
data = Dict()
data["N"] = N
data["x"] = x
data["mu1"] = 2
data["mu2"] = 5

# Set up Dirichlet prior and mixture model likelihood
prior = NamedTupleDist(
    θ = Dirichlet([3, 1]),
);

likelihood = let d = data
    
    params -> begin
        
        ll_value = 0
        for i in 1:d["N"]
            comp1 = params.θ[1] * pdf(Normal(d["mu1"], 1), d["x"][i])
            comp2 = params.θ[2] * pdf(Normal(d["mu2"], 1), d["x"][i])
            ll_value += log(comp1 + comp2)

        end
        
        return LogDVal(ll_value)
    end
end

# Posterior
posterior = PosteriorDensity(likelihood, prior);

# Sample
samples = bat_sample(posterior, MCMCSampling(mcalg=MetropolisHastings())).result
```

PR #275, opened 2021-04-13
List of algorithms in docu

Cornelius-G:
We have been asked multiple times now if we could provide a list of algorithms included in BAT.jl.
I guess such a list would also be nice for advertising the tool, so I just started trying to list them. 
I'm not sure where exactly we would want to put this. For now I just made a new page in the docs for it.
I would also propose to include short examples for each algorithm, so that this list can be referred to when trying to use the different algorithms.


docs/src/list_of_algorithms.md
```diff
@@ -0,0 +1,60 @@
...
+
+-   [Adaptive Harmonic Mean Integration (AHMI)](https://bat.github.io/BAT.jl/dev/stable_api/#BAT.AHMIntegration)
+
+-   [Vegas Integration](https://bat.github.io/BAT.jl/dev/experimental_api/#BAT.VEGASIntegration)
+
+-   [Cuhre Integration](https://bat.github.io/BAT.jl/dev/experimental_api/#BAT.CuhreIntegration)
+
+-   [Divonne Integration](https://bat.github.io/BAT.jl/dev/experimental_api/#BAT.DivonneIntegration)
+
+## Optimization algorithms:
```
oschulz, line 56: Let's name this "Mode finding algorithms"?

docs/src/list_of_algorithms.md
```diff
@@ -0,0 +1,60 @@
...
+	```
+	MCMCSampling(mcalg = HamiltonianMC())
+	```
+
+-   [Reactive Nested Sampling](https://bat.github.io/BAT.jl/dev/experimental_api/#BAT.ReactiveNestedSampling)
+	```
+	ReactiveNestedSampling()
+	```
+
+-   [Partitioned Sampling](https://bat.github.io/BAT.jl/dev/experimental_api/#BAT.PartitionedSampling)
```
oschulz, line 28: Let's use `[Partitioned Sampling (experimental)]` for now.

docs/src/list_of_algorithms.md
```diff
@@ -0,0 +1,60 @@
...
+
+-   [Vegas Integration](https://bat.github.io/BAT.jl/dev/experimental_api/#BAT.VEGASIntegration)
+
+-   [Cuhre Integration](https://bat.github.io/BAT.jl/dev/experimental_api/#BAT.CuhreIntegration)
+
+-   [Divonne Integration](https://bat.github.io/BAT.jl/dev/experimental_api/#BAT.DivonneIntegration)
+
+## Optimization algorithms:
+
+-  [Nelder-Mead algorithm](https://bat.github.io/BAT.jl/dev/stable_api/#BAT.MaxDensityNelderMead)
```
oschulz, line 58: Let's call it `[Nelder-Mead Optimization]`?

docs/src/list_of_algorithms.md
```diff
@@ -0,0 +1,60 @@
...
+
+-   [LBFGS](https://bat.github.io/BAT.jl/dev/stable_api/#BAT.MaxDensityLBFGS)
```
oschulz, line 60: Let's call it `[LBFGS Optimization]`?

docs/src/list_of_algorithms.md
```diff
@@ -0,0 +1,60 @@
+
+# List of algorithms 
+
+BAT.jl offers performant implementations of mutliple algorithms for sampling, integration and optimization.
+
+## Sampling algorithms
+
+- [IIDSampling](https://bat.github.io/BAT.jl/dev/stable_api/#BAT.IIDSampling)
+	```
```
oschulz, line 9: Doesn't need to be indented, headers are used instead of list items. Also, let's always use "```julia" in the opening triple-backquote.

docs/src/list_of_algorithms.md
```diff
@@ -0,0 +1,60 @@
...
+	IIDSampling(nsamples=10^5)
```
oschulz, line 10: Maybe

```julia
smpls = IIDSampling(nsamples=10^5).result
```

or so, to be a bit more illustrative? And then same same for the others?

docs/src/list_of_algorithms.md
```diff
@@ -0,0 +1,60 @@
+
+# List of algorithms 
+
+BAT.jl offers performant implementations of mutliple algorithms for sampling, integration and optimization.
+
+## Sampling algorithms
+
+- [IIDSampling](https://bat.github.io/BAT.jl/dev/stable_api/#BAT.IIDSampling)
```
oschulz, line 8: Let's use headers for the algorithms, instead of list items. Also, shouldn't use full URLs to link to other parts of the docs, it's fragile (and would link from dev docs to stable docs). Maybe something like this?

### IIDSampling

BAT algorithm type: [`IIDSampling`](@ref)

Example:

```julia
smpls = IIDSampling(nsamples=10^5).result
```

oschulz (changes requested):
Thanks a lot for taking the initiative here, it's really high time we add this.

oschulz:
Thanks! Good to merge from by side, as soon as the conflicts are resolved.

oschulz:
Replaced by #294

PR #278, opened 2021-05-01
[WIP] Partitioned sampling

jlrestrepol:
Hi @oschulz, Hi @VasylHafych 

As we discussed I'm creating this PR to discuss the partitioned sampling test and when everything is done to request to merge, I'm sorry that it's a bit long. A few points:

1) Exploration Sampler: Oliver suggested using Sobol sampler, it's very fast but sometimes(2 or 3 times out of 20) two modes end up in the same subspace which can screw up the KS test, here I called mcmc (and commented out Sobol) with only 400 steps and it always worked well. The performance loss is maybe 5 seconds. It's off course up to you but my suggestion would be to leave mcmc. 

2) Integrator: Oliver suggested using cuba, however I don't seem to be able to make it work because cuba-based integration algorithms in BAT (I think) need an abstract density and Vasyl performs the integration  with samples drawn from the density rather than the density itself. This is the code from `sample_subace`: `integras_subspace = bat_integrate(samples_subspace, integration_algorithm).result`. Maybe I'm missing something here?

3) testset "Partitioned Sampling"there I'm checking that the modes fall in different subspaces, Oliver, you said there was a BAT function to check if a point lied in a hypervolume, I couldn't find it, could you maybe give me the name? For the moment I'm simply going through the matrix and doing the check by hand.

4) potentially create a testset called "Sampling Subspaces" where I retrieve the samples from each subspace and check that they actually resemble the individual gaussians, I couldn't find a way to retrieve those samples, is there a way to do it Vasyl?


oschulz:
> there was a BAT function to check if a point lied in a hypervolume, I couldn't find it, could you maybe give me the name?

https://github.com/bat/BAT.jl/blob/6a75b0ed6d714422b076c0d26e7d00ac9d4e91ee/src/variates/spatialvolume.jl#L98

oschulz:
> using cuba, however I don't seem to be able to make it work because cuba-based integration algorithms in BAT (I think) need an abstract density

Ah, good point. It should definitely be possible to use integrators that need the density, as well as integrators that need samples (some future ones may need both). We'll probably need to turn `SampledDensity` into a fully-fledged `AbstractDensity` first. If you use AHMI for now, can you choose the settings so that it runs fairly quicky? Maybe @VasylHafych  can advise.

jlrestrepol:
Thank you for your comments and specially for forwarding me the function and I'm sorry I missed it. I'll work on them during the week.

oschulz:
No worries!

jlrestrepol:
Hi @oschulz, hi @VasylHafych 

1) Sobol as exploration sampler now works, I simply changed the prior and now the modes are separated in subspaces: ran the test 50 times and the worst p_value I got was 0.86, set the threshold to 0.7 to be safe.

2) Integrator: I tried out a few different parameters for AHMI and chose those that worked best. Regarding your comment about how to use Cuba-based algorithms let me know if I could help somehow. 

3) I changed the implementation to use the function you sent me

4)  I guess is not need because we have the KS test

5) Results of benchamrking the current test


BenchmarkTools.Trial: 
memory estimate:  3.34 GiB
allocs estimate:  37194580

minimum time:     18.025 s (4.67% GC)
median time:      18.025 s (4.67% GC)
mean time:        18.025 s (4.67% GC)
maximum time:     18.025 s (4.67% GC)

samples:          1
evals/sample:     1

To compile and run takes 1:10 mins approx. is this acceptable?

jlrestrepol:
Also, idk why the Codevco report says we're decreasing the coverage, I double checked and I'm up to date with `master`, should I do something about this?

oschulz:
> Also, idk why the Codevco report says we're decreasing the coverage

Yes, that's just strange, sometimes.

oschulz:
> Sobol as exploration sampler now works, [...]

Very nice. @VasylHafych , could you take a look and let me know if this is good to merge from your side?

VasylHafych:
Hi @jlrestrepol,

> Very nice. @VasylHafych , could you take a look and let me know if this is good to merge from your side?

I had a quick look at your tests, they all seem to be reasonable to me. 

Regarding CUBA-based integrators, I haven't tested them with space partitioning yet. I might have a look at it in the coming weeks.

jlrestrepol:
Thanks @VasylHafych for your comments. 

I improved the comments because there were some questions there and a few things that could be explained better. 

@oschulz should I remove the WIP from the PR title?

oschulz:
>  should I remove the WIP from the PR title?

Yes, looks ready. Just add a newline at the end of "test_partitioned_sampling.jl" please.

jlrestrepol:
Done!

oschulz:
Thanks!

jlrestrepol:
I was checking the Code Coverage and it seems to me that the partitioned sampling part is till in red, should I do something about it? @oschulz

oschulz:
Yes, coverage calculation is a bit weird sometimes, I'll have to take a look.

PR #283, opened 2021-05-15
Adds tests for convolutions

jlrestrepol:
Hi @oschulz 

Taking into account what we discussed yesterday that I should not test `valuesahpes_utils.jl`, I think there's only need to test `convolutions.jl` in the utils library and this is what I do. In Codevco it appears that `array_utils.jl` is not tested but there's in fact a  test script called `test_array_utils.jl` for this module which runs and has tests for it. Let me know if you want me to test this further or if you have any general comments.

### DIFF
--- a/test/utils/test_convolutions.jl
+++ b/test/utils/test_convolutions.jl
@@ -1,0 +1,44 @@
+# This file is a part of BAT.jl, licensed under the MIT License (MIT).
+
+using BAT
+using Test
+using Distributions
+
+#2d Convolution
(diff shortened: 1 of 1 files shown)


oschulz:
Ok, that's fine, then.

jlrestrepol:
I'll be sure to start with the multivariate distributions next week then.

jlrestrepol:
Hi @oschulz,

Should I do something else in this PR?

oschulz:
This is fine, thanks!

jlrestrepol:
All right!

PR #289, opened 2021-05-28
Tests multimodal distributions

jlrestrepol:
Hi @oschulz,

This is the work I've done in Multivariate Distributions. I've a few questions/comments I'ld like to ask before requesting to merge:

1) As before, Codevco shows a misleading test coverage of the library, I only added test for the modules that didn't have before(funnel, gaussian shell, multimodal Cauchy) even though they look like they're tested in the Codvevco plot. 
2) what kind of tests would you like to see for the `rand` method implemented on the distributions? An example of such a method would be `Distributions.rand(rng::AbstractRNG, d::GaussianShell) = vec(rand(rng, d, 1))`. Should I check that those random numbers are uncorrelated and that they follow the parent distribution using a Kolmogorov-Smirnov test? 
3) I got a few weird errors on Multimodal Cauchy: all three `mean`, `cov`, `var` with the default parameters and with custom ones return `NaN`s, is this expected? 

- @test @inferred(Distributions._logpdf(mmc, [-1., 0., 1., 2.])) fails inferring the type and Distributions._logpdf(mmc, [-1., 0., 1., 2.]) succeeds. This looks a lot like a newbie mistake, but I dont know what am I doing wrong, sorry.


oschulz:
> I got a few weird errors on Multimodal Cauchy [...]

`mean` should return an sensible result. Cauchy dists have infinite variance.

We should consider replacing the multimodal cauchy with a multimodal T-distribution (which would be the same for alpha = 1, but more flexible).

oschulz:
> fails inferring the type and Distributions._logpdf

Must be some type instability in the code. @sthayashi has become good at hunting them down, maybe he can assist.

sthayashi:
> > fails inferring the type and Distributions._logpdf
> 
> Must be some type instability in the code. @sthayashi has become good at hunting them down, maybe he can assist.

Is this type instability only when `Distributions._logpdf` is called? Or does it also have this problem for `Distributions.logpdf`?

jlrestrepol:
Thanks for the quick reply, unfortunately I'm busy with another already and I can't answer right now. I'll get back to you later.

sthayashi:
I made a quick PR #290 that fixes this problem. Note it doesn't give any performance gains, because the type instability comes from the way that `Distributions._logpdf` sums the elements of array `x`.

oschulz:
> I made a quick PR #290 that fixes this problem.

Merged, thanks!

jlrestrepol:
> I made a quick PR #290 that fixes this problem. Note it doesn't give any performance gains, because the type instability comes from the way that `Distributions._logpdf` sums the elements of array `x`.

Thanks to @sthayashi  for the PR, it fixes the bug! :)





> > I got a few weird errors on Multimodal Cauchy [...]
> 
> `mean` should return an sensible result. Cauchy dists have infinite variance.
> 
> We should consider replacing the multimodal cauchy with a multimodal T-distribution (which would be the same for alpha = 1, but more flexible).

@oschulz  I'm a little bit confused then, I calculated the sample mean 10 times for 10.000 random samples and doesn't look like I'm getting a consistent result:

`for i in 1:10
           println(mean(rand(BAT.MultimodalCauchy(), 10^4)))
       end ` gives : `
-0.22261536087117348
-0.06653815090285499
-0.2359563850221311
0.32563289042591953
-0.19034549463015077
0.5934563377424266
0.7369440653552619
0.5615497648882243
1.2775215912060776
-1.9677966995685405`

Also, do you want me to write some (probably naive) implementation of the multimodal t-distribution to test it?

oschulz:
Due to the way `rand` works for multivariate distributions, `rand(BAT.MultimodalCauchy(), 10^4)` returns a matrix, not a vector of vectors. You have to do the mean along the first dimension.

But you're right, of course, the mean of a Cauchy distribution is not defined as well.

oschulz:
> Also, do you want me to write some (probably naive) implementation of the multimodal t-distribution to test it?

Yes, why not, have a try.

jlrestrepol:
> Due to the way `rand` works for multivariate distributions, `rand(BAT.MultimodalCauchy(), 10^4)` returns a matrix, not a vector of vectors. You have to do the mean along the first dimension.

Oh, I see, thanks!

> 
> But you're right, of course, the mean of a Cauchy distribution is not defined as well.

Then just to be on the same page: mean, cov and var are not defined for multivariate Cauchy, the type instability was fixed in #290 and what's pending is to test the `rand` method with KS or similar, then ask to merge this PR and then do some prototyping of the multivariate t-distribution, are you ok with that plan @oschulz ?

oschulz:
> are you ok with that plan @oschulz

Sounds good!

jlrestrepol:
Ok, then I'll work on that on Friday the whole day.

jlrestrepol:
@oschulz so after reading I think the best test for checking the random numbers extracted from our multivariate distributions is One Sample KS test, but afaik it's only for univariate distributions and further BAT has only the Two Sample KS Test implemented. I can test the distributions  in only 1d using an [external package](https://juliastats.org/HypothesisTests.jl/stable/nonparametric/#Kolmogorov-Smirnov-test-1) but I would have to basically convert the distributions from multivariate to univariate which I'm not sure it's useful because this library is supposed to be multivariate. Should I do that or do you have a better idea?

jlrestrepol:
Hi @oschulz, 

So with the help of @sthayashi I implemented the KS tests for the Funnel, Gaussian Shell and Multimodal Cauchy distributions in limit cases. Can you have a look and let me know what you think? From my side, this is ready to merge.

oschulz:
What's the runtime of these tests, approximately?

jlrestrepol:
These are the results using Benchmarktools

**Funnel:**

BenchmarkTools.Trial: 
  memory estimate:  972.99 MiB
  allocs estimate:  8394

  minimum time:     5.394 s (6.31% GC)
  median time:      5.394 s (6.31% GC)
  mean time:        5.394 s (6.31% GC)
  maximum time:     5.394 s (6.31% GC)
  
  samples:          1
  evals/sample:     1

**Gaussian Shell:**

BenchmarkTools.Trial: 
  memory estimate:  696.09 MiB
  allocs estimate:  31941481

  minimum time:     3.485 s (2.26% GC)
  median time:      3.509 s (2.18% GC)
  mean time:        3.509 s (2.18% GC)
  maximum time:     3.532 s (2.10% GC)

  samples:          2
  evals/sample:     1

**Multimodal Cauchy:**

BenchmarkTools.Trial: 
  memory estimate:  3.05 GiB
  allocs estimate:  80006544

  minimum time:     16.485 s (7.54% GC)
  median time:      16.485 s (7.54% GC)
  mean time:        16.485 s (7.54% GC)
  maximum time:     16.485 s (7.54% GC)

  samples:          1
  evals/sample:     1

jlrestrepol:
Multimodal Cauchy is the slowest one basically because I perform the KS two times, one in each dimension. 2 is the minimum number of dimension it can have. Should I take one of the tests out?

oschulz:
Can you just reduce the number of samples?

jlrestrepol:
Ran both tests 50 times and there wasn't any test failures with `n_samples=10^6`. The running time is one order of magnitud less.

oschulz:
Nice!

PR #296, opened 2021-06-18
Space partitioning - Resampling Update 

VasylHafych:
Hi @oschulz, @jlrestrepol, 

As we discussed the last Friday, I have created a pull request with upgraded space partitioning sampling. I have tested it on simple problems with up to 5 dimensions, and the current version works well.

Oliver suggested that we implement space transformation to make the algorithm more efficient. @jlrestrepol, would you be able to help me with this part? 

If so, we can discuss below the concrete steps that need to be done, together with the test suite to check the accuracy of the algorithm. 

Thanks and cheers


VasylHafych:
@oschulz, how should we deal with the convergence criteria for each subspace? I have this code that works for `BrooksGelmanConvergence`. It converts the resulting `DensitySampleVector` into a vector of chains that are later used for convergence check: 
```julia 
function check_convergence(ct::BrooksGelmanConvergence, samples::DensitySampleVector)
    
    # create a vector of chains
    chains_ind = unique([i.chainid for i in samples.info])
    vector_chains = DensitySampleVector[]
    for i in chains_ind
        mask_chain = [j.chainid == i for j in samples.info]
        push!(vector_chains, samples[mask_chain])
    end
    
    # apply convergence 
    res = bg_R_2sqr(MCMCBasicStats.(vector_chains), corrected = ct.corrected)
    
    max_Rsqr = maximum(res)
    converged = max_Rsqr <= ct.threshold
    @debug begin
        success_str = converged ? "have" : "have *not*"
        "Chains $success_str converged, max(R^2) = $(max_Rsqr), threshold = $(ct.threshold)"
    end
   GRConvergenceResult(converged, max_Rsqr)
    
end
```


But how to be with the samples that have no chains?

jlrestrepol:
Hi @VasylHafych,

Yes, off course, happy to work on it next week.

VasylHafych:
@jlrestrepol, great, thanks. 

To progress with the work, we need a small help from @oschulz. Namely, we transform the original space to the uniform by using `PriorToUniform`. After this, the binary tree defines rectangular subspaces in the transformed uniform space. To generate samples in each subspace, we need to truncate/constrain the sampler to the subspace boundaries. In the old version, this was done by using the `truncate_density` function. Now, when space is uniform, that function is no longer working, and it should be extended to `StandardMvUniform` distributions. I have tried this simple implementation, but it does not work: 
```julia
function truncate_density(density::AbstractPosteriorDensity, bounds::AbstractVarBounds)
    old_prior = getprior(density)
    old_bounds = var_bounds(old_prior)
    new_bounds = ismissing(old_bounds) ? bounds : var_bounds(old_prior) ∩ bounds
    PosteriorDensity(density.likelihood, DistributionDensity(density.prior.dist, new_bounds), new_bounds, density.parshapes)
end
```
Oliver, can you help us to define the `truncate_density` function, where the prior is given by `StandardMvUniform`? You have also mentioned that we should replace `TruncatedDensity` with `RenormalizedDensity`, removing the `bounds` field.

oschulz:
> To progress with the work, we need a small help from @oschulz.

On the way: #297

VasylHafych:
Hi Oliver! I just checked your updates on a few simple examples — code seems to work well. Thanks for this help. 

The next question in my list is how to evaluate the convergence test in each subspace.  Different samplers use different convergence tests, so we need to agree on some general way how to perform it.

oschulz:
>  evaluate the convergence test in each subspace

We should add an `isvalid` field to the output of `bat_sample`.

VasylHafych:
I assume you suggest that we apply a convergence test to the final sample and not use the test result from the tuning phase? 

I have done this for the Metropolis sampler, but I need some help with HMC, NS, Sobol, etc. (especially those from the BAT external packages). I never dealt with their convergence criteria, so I am not familiar with how they work. Also, this should probably be a separate pull request.

oschulz:
> I assume you suggest that we apply a convergence test to the final sample and not use the test result from the tuning phase?

Maybe we should add an option to the samplers to do that automatically (if requested by the user, or maybe even by default).

VasylHafych:
Hi @oschulz. 

I have implemented all of the changes from my side, and this pull request is ready to be reviewed/merged. I used this code to generate an example that we included in the space partitioning paper. It is now possible to automatically repartition subspaces where the convergence criteria were not successful. 

Regarding the convergence test, I defined it only for MCMC sampling. You can ask maintainers of other samplers to add their own `isvalid` fields to `bat_sample`. Until then, this sampling approach will work only with MCMC. 

After you check and merge this PR, we can ask @jlrestrepol to perform a few benchmarks runs with different target densities.

oschulz:
Nice!

PR #298, opened 2021-06-24
ahmc type stability fixes

sthayashi:


sthayashi:
https://github.com/sthayashi/BAT.jl/blob/8d5f97be291ebd570ce52ec6577272a5ad849b7c/src/samplers/mcmc/ahmc/ahmc_sampler.jl#L116-L117 is not type stable for me. Haven't found a good solution for it.

oschulz:
> ... is not type stable for me

Yep, we'll have to take another look at that one day - but it's during chain creation, so it shouldn't affect chain iteration.

oschulz:
Thanks!

PR #300, opened 2021-07-02
[WIP] Tests for modified partitioned sampling

jlrestrepol:
Hi @VasylHafych 

Maybe we can discuss here the tests for your extension of partition sampling. As far as I understood you added two new features: re-sampling of subspaces when the convergence is not reached (up to a certain number of times spececified by `nmax_resampling`) and the transformation of the prior to make it uniform. I modified a bit the testing script I had to include a non-uniform prior, a Gaussian and changed the algorithms a bit. Is this in the direction you want? A few comments:

* There might be some type instability in line 34 that makes @inferred fail in bat_sample, test fails if I leave it.
*  I got an error while using Sobol sampler when resampling subspaces. The exception said Sobol didn't have a `nstep`  parameter so it looks like resampling only accepts mcmc for the moment, right? I changed exploration alg to mcmc and worked
*  For the histogram comparison you said in the email we should compare iid to space partitioning, is this preferred above mcmc vs space partitioning? I  implemented both and simply commented mcmc out to leave iid. I understood that's what you wanted, let me know if that's not the case.

Please let me know of any general comments you have.

### DIFF
--- a/test/samplers/partitioned_sampling/test_partitioned_sampling.jl
+++ b/test/samplers/partitioned_sampling/test_partitioned_sampling.jl
@@ -9,7 +9,6 @@ using Distributions, LinearAlgebra, ValueShapes
     σ = [0.1 0 0; 0 0.1 0; 0 0 0.1]
     μ = [1 1 1; -1 1 0; -1 -1 -1; 1 -1 0]
     mixture_model = MixtureModel(MvNormal[MvNormal(μ[i,:], Matrix(Hermitian(σ)) ) for i in 1:4])
-    # ToDo: Use more complex prior with non-uniform distribution
     prior = NamedTupleDist(a = [Normal(0,2), Normal(0,2), Normal(0, 2)])
     likelihood = let model = mixture_model
         params -> LogDVal(logpdf(model, params.a))
@@ -21,8 +20,7 @@ using Distributions, LinearAlgebra, ValueShapes
     #Sampling and integration algorithms
     mcmc = MCMCSampling(mcalg = MetropolisHastings(), nsteps = 10^3);
     ahmi = ahmi = AHMIntegration(whitening = BAT.NoWhitening(), max_startingIDs = 10^3)
-    #sobol = BAT.SobolSampler(nsamples = 500)
-    mcmc_exp = MCMCSampling(mcalg = MetropolisHastings(), nsteps = 1000, nchains=20, strict=false)
+    mcmc_exp = MCMCSampling(mcalg = MetropolisHastings(), nsteps = 400, nchains=4, strict=false)
 
     ps = PartitionedSampling(sampler = mcmc, npartitions=4, exploration_sampler=mcmc_exp, integrator = ahmi, nmax_resampling=5);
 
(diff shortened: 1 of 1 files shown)


VasylHafych:
Hi @jlrestrepol,

Thanks for this PR. That is exactly what we want to have. 

- The Sobol sampler will fail with the Space Partitioning because we have not yet implemented a convergence check. We should test it only with the MH sampler. 
- IID is definitely better because it is not correlated and more representative. I would use IID if your target density allows for it. 
- Regarding the type instability: Was this also the case for the previous version of the algorithm? It would be great if you could spot what leads to this problem inside of `bat_sample`. I can then try to fix it.

jlrestrepol:
> Hi @jlrestrepol,
> 
> Thanks for this PR. That is exactly what we want to have.
> 
Great! 
> * The Sobol sampler will fail with the Space Partitioning because we have not yet implemented a convergence check. We should test it only with the MH sampler.
All right.
> * IID is definitely better because it is not correlated and more representative. I would use IID if your target density allows for it.
Perfect then
> * Regarding the type instability: Was this also the case for the previous version of the algorithm? It would be great if you could spot what leads to this problem inside of `bat_sample`. I can then try to fix it.
I haven't done it before, but I'm happy to have a look.

jlrestrepol:
@oschulz looks like Vasyl is happy with the current state of the tests for the extended partitioned sampling algorithm. Do you have any general comments? The output of `@benchmark` is:

BenchmarkTools.Trial: 
  memory estimate:  20.70 GiB
  allocs estimate:  107301575

  minimum time:     46.485 s (11.10% GC)
  median time:      46.485 s (11.10% GC)
  mean time:        46.485 s (11.10% GC)
  maximum time:     46.485 s (11.10% GC)

  samples:          1
  evals/sample:     1

If not then I'll clean up the code and request to merge.

oschulz:
1 minute testing time just for the partitioned sampling is kinda heavy, but let's leave it like that for now if there's no other way.

test/samplers/partitioned_sampling/test_partitioned_sampling.jl
```diff
@@ -20,20 +20,22 @@ using Distributions, LinearAlgebra, ValueShapes
 
     #Sampling and integration algorithms
     mcmc = MCMCSampling(mcalg = MetropolisHastings(), nsteps = 10^3);
-    #vegas = BAT.VEGASIntegration(trafo = NoDensityTransform(), rtol = 0.001, atol = 1.0e-8)
     ahmi = ahmi = AHMIntegration(whitening = BAT.NoWhitening(), max_startingIDs = 10^3)
-    sobol = BAT.SobolSampler(nsamples = 500)
-    #mcmc_exp = MCMCSampling(mcalg = MetropolisHastings(), nchains =4, nsteps = 400, trafo = NoDensityTransform(),);
+    #sobol = BAT.SobolSampler(nsamples = 500)
+    mcmc_exp = MCMCSampling(mcalg = MetropolisHastings(), nsteps = 1000, nchains=20, strict=false)
```
VasylHafych, line 25: I think the test should be much faster if you use `nsteps` 300-500. Can you try this?

jlrestrepol: Done!

jlrestrepol:
So with Vasyl's suggestion now it takes approximately 30 seconds, this is the output of benchmarktools

BAT_partitioned_sampling |    5      5
BenchmarkTools.Trial: 
  memory estimate:  4.85 GiB
  allocs estimate:  90002880

  minimum time:     28.588 s (5.37% GC)
  median time:      28.588 s (5.37% GC)
  mean time:        28.588 s (5.37% GC)
  maximum time:     28.588 s (5.37% GC)


  samples:          1
  evals/sample:     1

@oschulz what do you think?

oschulz:
I think let's merge it. :-)

PR #303, opened 2021-07-10
Error correction and design update of bridge sampler

celaue:
Hi @oschulz,

this commit corrects the error estimate and simplifies parts of my older implementation.

Thanks,

Cedrik


oschulz:
Thanks!

PR #305, opened 2021-07-15
Multimodal t distribution

jlrestrepol:
Hi @oschulz,

As we discussed I'm creating this PR that implements a multimodal student-t distribution following closely what you did in the multimodal Cauchy module. 

1) The functional form is equivalent to the definition of multimodal Cauchy in the paper

Screenshot 2021-07-15 at 15 56 21

2) For d (degrees of freedom) = 1 you get back the multimodal Cauchy as it's shown by the results of the KS test:

mmc = BAT.MultimodalCauchy(μ = 60., σ = 1., n=2);
mt = BAT.MultimodalStudentT(μ=60, σ=1., d = 1, n = 2);

ApproximateTwoSampleKSTest(rand(mt, 10^4)[2,:], rand(mmc, 10^4)[2,:])
Approximate two sample Kolmogorov-Smirnov test

Population details:
    parameter of interest:   Supremum of CDF differences
    value under h_0:         0.0
    point estimate:          0.0079

Test summary:
    outcome with 95% confidence: fail to reject h_0
    two-sided p-value:           0.9139

Details:
    number of observations:   [10000,10000]
    KS-statistic:              0.5586143571373726

3) This is a Plot of the distribution I get for d = 1. Looks fine at least to me.

Screenshot 2021-07-15 at 16 07 37

4) The mean, cov and var are undefined for d = 1 as expected, but give finite results for, e.g., d = 2

`mt = BAT.MultimodalStudentT(μ=60, σ=1., d = 3, n = 2);`
`Statistics.mean(mt) = [0.0, 0.0]`
`Statistics.cov(mt) = [3603.0 0 ; 0 3603.0]` 

5) Following `multimodal_cauchy.jl` I overwrite `_construct_dist` in this way: `function _construct_dist(mixture_model::MixtureModel, σ::Real, d::Int, n::Integer)`  I feel like the signature `MixtureModel`is not specific enough and I would like to make it something like `MixtureModel{TDist}` but I haven't been able to do it, is there a way to do that?

Maybe let's discuss here any changes you may want and after that, I also include some tests?


jlrestrepol:
Hi @VasylHafych, this is the PR we talked about during the meeting.

jlrestrepol:
Hi @oschulz,

I changed the implementation of MultimodalCauchy so it calls internally MultimodalStudentT as you suggested last Friday. I'm unsure, though, if my approach was the correct way of doing it. Basically, I do this:

Screenshot 2021-07-23 at 10 30 04


I also kept all the other functions in the module `multimodal_cauchy.jl` instead of inheriting them from `multimodal_student-t.jl`. I did that to keep the datatype `MultimodalStudentT` functional which I guess is your intention. 

I also made sure that the implementation was working for `n>2`.

Is this in the direction you want?

Please let me know what do you think.

oschulz:
Hm, on further reflection, better just remove `MultimodalCauchy` completely, and replace it with the equivalent `MultimodalStudentT` in the tests and benchmarks.

jlrestrepol:
All right! I'll do this and extend the testing to include the new functionalities this week.

jlrestrepol:
Hi @oschulz,

Can you have a look? I implemented the changes we talked about.

jlrestrepol:
Apparently, there's an error in Documentation, the traceback says:

`Warning: undefined binding 'BAT.MultimodalCauchy' in `@docs` block in src/experimental_api.md:11-41`

and also

`┌ Error: 1 docstring not included in the manual:`
` │ ` 
` │     BAT.MultimodalStudentT`

I couldn't find the file `src/experimental_api.md:11-41` so I guess it's not part of the public Github repo, right? How could I tr to fix this issue?

oschulz:
> I guess it's not part of the public Github repo, right? 

Yes it is :-)

https://github.com/bat/BAT.jl/blob/master/docs/src/experimental_api.md

jlrestrepol:
> > I guess it's not part of the public Github repo, right?
> 
> Yes it is :-)
> 
> https://github.com/bat/BAT.jl/blob/master/docs/src/experimental_api.md

Oh, I see, I was looking for it in `/src` thanks for sharing the link!

jlrestrepol:
Hi @oschulz,

Can you have a look and let me know if there's something else you want me to do here?

oschulz:
`MultimodalCauchy` is also used in

./examples/benchmarks/functions_1D.jl
./examples/benchmarks/run_benchmark_ND.jl
./examples/benchmarks/functions_2D.jl

we need to replace those by `MultimodalStudentT` as well. Afterwards, you should run "benchmarks.jl" from within the "examples/benchmarks" directory (will take 4 hours or so) and see if runs through without error.

jlrestrepol:
Hi @oschulz,

I implemented the changes. It's important to note that `MultimodalStudentT` and `MultimodalCauchy` can't be run with `dim=1`, they're at least two dimensional. For this reason, I commented out `multimodal_cauchy` in the dictionary `testfunctions_1D` in `functions_1D.jl`

examples/benchmarks/functions_1D.jl
```diff
@@ -6,7 +6,7 @@ cauchy = (posterior=Cauchy(0,1),mode=0,mean=0,var=Inf,chi2=[9999],ks=[9999.],ahm
 #cauchy_variance_for_edges = x -> 0.6365969042714238*x+-0.9834816540023621
 
 funnel = (posterior=BAT.FunnelDistribution(1.2,0.,1),mode=0,mean=0,var=2.07350,chi2=[9999],ks=[9999.],ahmi=[9999.])
-multi_cauchy = (posterior=BAT.MultimodalCauchy(µ=5.,σ=2.,n=1),mode=[-5,5],mean=0,var=Inf,chi2=[9999],ks=[9999.],ahmi=[9999.])
+multi_cauchy = (posterior=BAT.MultimodalStudentT(µ=5.,σ=2.,d=1,n=2),mode=[-5,5],mean=0,var=Inf,chi2=[9999],ks=[9999.],ahmi=[9999.])
```
oschulz, line 9: We shouldn't have a 2-dim dist in the list of 1D test functions. But maybe we could a univariate multi-modal cauchy dist to our test distributions?

oschulz (changes requested):

examples/benchmarks/functions_1D.jl
jlrestrepol: Yes, I can do this if that's what you want. Just want to point out that this is explicitly against the definition of multimodal Cauchy given in the paper(n >=2) and contrary to the code we had for multimodal Cauchy (a similar line was also commented for this case). Should I do it anyway?

oschulz: We should do it as a separate univariate distribution.

jlrestrepol: How about marginalizing the 2D posterior or calling a lambda function? Honestly, seems to me that creating a 1D function with the same code but without the `argcheck n > 1`  is forcing it too much.

oschulz: Just take it out of the 1D test functions then, we can worry about it some other time.

jlrestrepol:
Hi @oschulz,

I implemented the changes and ran the tests again (took >6 hours ) without problems, can you have a look? The automatic checks failed on CI/Julia 1 in Ubuntu(files I didn't touch) but succeeded in Mac, I guess is a p-value or statistical thing, what do you think?

oschulz:
> I guess is a p-value or statistical thing, what do you think

Looks like it.

Thanks, I'll merge!

PR #309, opened 2021-07-29
Fix problems due to new Trajectory type in AdvancedHMC.jl

sthayashi:


sthayashi:
`BAT` currently only implements the NUTS sampler from `AdvancedHMC.jl`. Thus the previous commit https://github.com/bat/BAT.jl/commit/bb00ab4357535a21e37a4df00ce32a1f7722a1a7 only made the necessary changes to allow the use of NUTS to resume. This PR brings `BAT` back to the previous problem with non-NUTS samplers, namely it assumes there is a tree depth, which a static trajectory does not have.

src/samplers/mcmc/ahmc/ahmc_sampler.jl
```diff
@@ -73,6 +73,9 @@ mutable struct AHMCIterator{
     proposal::PL
 end
 
+function AdvancedHMC.transition(rng::AbstractRNG, hamiltonian::AdvancedHMC.Hamiltonian, proposal::AdvancedHMC.Trajectory, z::AdvancedHMC.PhasePoint)
```
oschulz, line 76: Can we do this differently? This is type piracy ...

oschulz (changes requested):

sthayashi:
Closing this PR. Master branch should already have the `ERROR: type NamedTuple has no field tree_depth` when using `BAT.FixedStepNumber` or `FixedTrajectoryLength` proposal.

PR #310, opened 2021-08-06
Tests for sampling with hmc

sthayashi:
Will add another test or two before ready for merge.


oschulz:
How does this relate to https://github.com/bat/BAT.jl/blob/master/test/samplers/mcmc/test_hmc.jl ?

sthayashi:
https://github.com/bat/BAT.jl/blob/master/test/samplers/mcmc/test_hmc.jl was originally to test all of HMC before it was known that only NUTS worked. I was planning to do three separate PRs: this one for testing NUTS, one for fixing the code to work with the other proposals, a final one to test the newly fixed proposals.

I think I can close the other PR if it's cluttering things up.

oschulz:
> I think I can close the other PR if it's cluttering things up.

Yes, let's merge that into one PR.

Can you append the additional tests to "test/samplers/mcmc/test_hmc.jl"?

sthayashi:
So just all of ahmc tests into one file, rather than a `test_ahmc_sampler.jl`, `test_ahmc_metric.jl`, etc.?

oschulz:
> So just all of ahmc tests into one file

Yep.

oschulz:
Unrelated test failures should be fixed now, can you merge with, resp. rebase on, master?

oschulz:
The remaining test failures should be relevant - we'll have to find a way to deal with these.

sthayashi:
Yes, this is exactly why I explored using the original `AdvancedHMC.jl` package to sample the same distribution and run statistical tests from its HMC sampler. It has the exact same problem of sporadically failing, which suggests this is a problem that may not be solvable directly within `BAT.jl`. I'm wondering how loose can the constraints on something like a KS/AD test can be, while still remaining meaningful. Obviously with enough of these sporadic failures throughout all of the tests in `BAT.jl`, the end result is that getting a failure is no longer what could be called sporadic.

oschulz:
We don't want to make the tests too loose - I guess the best thing might be to try repeating such test once if they fail. Loosening things a bit may be good though - I have an idea, let me try something out ...

oschulz:
The adaption should still work, but it's done via a tuner callback now, to decouple the chain from the tuner and allow for having different tuning algorithms:

https://github.com/bat/BAT.jl/blob/6d1e500cc32620a590e1f829721de43afa86e3c5/src/samplers/mcmc/mcmc_algorithm.jl#L247

https://github.com/bat/BAT.jl/blob/862d209f5555502e58cb0959661ec058876f14ab/src/samplers/mcmc/ahmc/ahmc_adaptor.jl#L136

sthayashi:
Since there have been a couple of merges with the master branch, the commit history is a lot longer. Better to reorder + squash commits? Or is it better to leave it as is for transparency during review?

oschulz:
> Better to reorder + squash commits? Or is it better to leave it as is for transparency during review?

We can just squash it at merge time.

oschulz:
> The commit history is pretty messy now. 

We could also squash it into a single commit at merge.

sthayashi:
Sorry for delaying this PR, but I think it should be ready after I squash everything. An amateur question: since there are a lot of merge conflicts, am I able to just squash all commit into one (even those not owned by me?), or does that ruin master branch's commit history? Otherwise I will just start a new PR with the relevant changes as a single commit.

oschulz:
I can just squash it on merge, don't worry about it. :-) Fine to merge from your side?

sthayashi:
Yes, should be fine to merge now. I rebuilt a couple times without any new test failures due to this PR (only the rare failure that already exists)

PR #323, opened 2021-10-28
Add tests for BAT.cov2pdmat() in test_distribution_functions.jl

Micki-D:
Wrote tests for the cov2pdmat() function in distributions_functions.jl

I'm unsure if the tests are appropriate and I also tried out doing a pull request for the first time here, so stuff might be garbled up somehow ;(

### DIFF
--- a/test/distributions/test_distribution_functions.jl
+++ b/test/distributions/test_distribution_functions.jl
@@ -4,7 +4,7 @@ using BAT
 using Test
 
 using LinearAlgebra, Random
-using Distributions, PDMats, StatsBase, SparseArrays
+using Distributions, PDMats, StatsBase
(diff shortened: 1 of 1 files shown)


Micki-D:
Alright, I'm sorry that this is such a hastle.
In the second method of cov2pdmat, the type of the input matrix is put as 'AbstractPDMat' but I can't test for that since you can't instantiate that type. That's why I tested for the subtypes.

oschulz:
>  can't instantiate that type. That's why I tested for the subtypes.

Oh sure! In such cases, using one or two common subtypes is usually fine.

oschulz:
Ok, the one test failure is just random.

PR #326, opened 2021-11-07
extended test coverage for log_uniform.jl

Micki-D:
I assumed the definition in line 33 of the src file:

LogUniform() = LogUniform{Float64}(0, 1)

 is supposed to catch an empty input as an arg error since the first arg of LogUniform needs to be finite per definition.

### DIFF
--- a/src/distributions/log_uniform.jl
+++ b/src/distributions/log_uniform.jl
@@ -30,7 +30,7 @@ function LogUniform(a::Real, b::Real)
     LogUniform{T}(a, b)
 end
 
-LogUniform() = LogUniform{Float64}(0, 1)
+LogUniform() = LogUniform{Float64}(1, 2)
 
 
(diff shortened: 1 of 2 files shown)


oschulz:
> `LogUniform() = LogUniform{Float64}(0, 1)` is supposed to catch an empty input

Ooops, no, it's not, it's just supposed to be a default ctor. Better change it to  `LogUniform() = LogUniform{Float64}(1, 2)`.

Micki-D:
Okeydoke, I changed the default ctor definition aswell as the test for it. 
The test will error though, since the master branch still has the old definition

oschulz:
> The test will error though, since the master branch still has the old definition

Just include the change in this PR.

Micki-D:
Yeah it already is

oschulz:
Nice!

PR #342, opened 2022-04-12
Fix `Error` in debug message

lmh91:
There was an error in some debug messages:
```julia
type HMCKernel has no field integrator
```


oschulz:
Thanks!

PR #346, opened 2022-04-28
Fix: don't use deterministic rng values for mode finding

Cornelius-G:
Fix 1: no longer use deterministic rng vlaues for initial values when using `bat_findmode`
Fix 2: use `_rand_v_for_target` when using `InitFromIID` and `InitFromSamples`

### DIFF
--- a/src/initvals/initvals.jl
+++ b/src/initvals/initvals.jl
@@ -137,11 +137,11 @@ export InitFromSamples
 
 
 function bat_initval_impl(rng::AbstractRNG, target::AnyDensityLike, algorithm::InitFromSamples)
-    (result = _rand_v(rng, target, algorithm.samples),)
+    (result = _rand_v_for_target(rng, target, algorithm.samples),)
 end
 
 function bat_initval_impl(rng::AbstractRNG, target::AnyDensityLike, n::Integer, algorithm::InitFromSamples)
-    (result = _rand_v(rng, target, algorithm.samples, n),)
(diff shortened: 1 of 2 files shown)


Cornelius-G:
> > Fix 1: no longer use deterministic rng vlaues for initial values when using bat_findmode
> 
> The problem is that we use reproducibility of results then, since `bat_findmode` doesn't have an RNG argument. Maybe it would be better to leave it the way it is and let users pass random starting points explicitly?

Ah... I found this quite non-intuitive and was actually wondering why our modefinder results were so consistent :D To me, it seems a bit like the opposite behavior as in `bat_sample`, so I was actually not expecting that. In `bat_sample` we expect different results and use a fixed `rng` if we want to reproduce. Especially, when explicitly passing an initval algorithm like `InitFromIID` to the optim algorithm, I expected it to draw randomly.  Is there a reason why we don't pass `rng`, similar to `bat_sample` ?

oschulz:
> Is there a reason why we don't pass rng, similar to bat_sample

Originally, the mode-finder's first algorithm was "return sample with max. posterior". So it seemed weird to have an RNG argument that wouldn't be used. But I agree, we could add an RNG argument - in fact we probably should.

Cornelius-G:
> But I agree, we could add an RNG argument - in fact we probably should.

Alright, I'm gonna take a look into adding an RNG argument to `bat_findmode` tomorrow

oschulz:
Thanks!

Cornelius-G:
Closed in favor of #347

PR #349, opened 2022-05-05
Import inverse-Hessian and WoodburyPDMat code from Pathfinder.jl

oschulz:
@sethaxen, is this Ok with you? Pathfinder.jl would add more than two seconds to the load time of BAT.jl, which is already way too long (we'll split it up soon, make things more modular, trying to get load time down). We just need `lbfgs_inverse_hessians` and `WoodburyPDMat` here since BAT already has mode-finding infrastructure and we'll add some new variational-inference code soon which will take are of ELBO calculation, etc.

After `WoodburyPDMat` is upstreamed to PDMats we'd remove it here, obviously, same for `lbfgs_inverse_hessians` if it does into some lower-level package.


sethaxen:
I'd prefer to try to figure out the source of the slow-down and handle it on the Pathfinder side, but if this is a feature you urgently need, then yes, this is in general fine. I do think it needs more thorough attribution consistent with the MIT license and ideally with a URL linking to the Pathfinder commit the code was taken from. Not certain the recommended way to do that.

RE load time, this is what I see on my machine:

On a fresh environment, Pathfinder might cause a 1.15s increase to BAT's loadtime
```julia
julia> @time using BAT
  7.783539 seconds (27.45 M allocations: 1.726 GiB, 6.28% gc time, 50.78% compilation time)

julia> @time using Pathfinder
  1.154934 seconds (4.32 M allocations: 289.421 MiB, 5.25% gc time, 52.73% compilation time)
```

On a new fresh environment, if we make these 4 dependencies optional for Pathfinder, it might only increase the loadtime by 0.2s. That's almost entirely due to code in PSIS.jl (necessary for multi-path Pathfinder; ~half of its load time after Distributions is due to PrettyTables though, and we could potentially drop that dependency) and Pathfinder.jl itself rather than their dependencies. Would that be enough of a speed-up to depend on Pathfinder itself?

```julia
julia> @time using BAT
  7.763856 seconds (27.45 M allocations: 1.726 GiB, 5.12% gc time, 50.26% compilation time)

julia> @time using AbstractDifferentiation
  0.401405 seconds (1.53 M allocations: 84.648 MiB, 2.92% gc time, 92.30% compilation time)

julia> @time using GalacticOptim
  0.378646 seconds (1.45 M allocations: 120.713 MiB, 18.09% compilation time)

julia> @time using Folds
  0.186595 seconds (692.59 k allocations: 38.897 MiB, 11.13% gc time, 50.81% compilation time)

julia> @time using Pathfinder
  0.193904 seconds (601.56 k allocations: 40.755 MiB, 16.95% compilation time)
```

sethaxen:
I'd prefer to try to figure out the source of the slow-down and handle it on the Pathfinder side, but if this is a feature you urgently need, then yes, this is in general fine. I do think it needs more thorough attribution consistent with the MIT license and ideally with a URL linking to the Pathfinder commit the code was taken from. Not certain the recommended way to do that.

RE load time, this is what I see on my machine:

On a fresh environment, Pathfinder might cause a 1.15s increase to BAT's loadtime
```julia
julia> @time using BAT
  7.783539 seconds (27.45 M allocations: 1.726 GiB, 6.28% gc time, 50.78% compilation time)

julia> @time using Pathfinder
  1.154934 seconds (4.32 M allocations: 289.421 MiB, 5.25% gc time, 52.73% compilation time)
```

On a new fresh environment, if we make these 4 dependencies optional for Pathfinder, it might only increase the loadtime by 0.2s. That's almost entirely due to code in PSIS.jl (necessary for multi-path Pathfinder; ~half of its load time after Distributions is due to PrettyTables though, and we could potentially drop that dependency) and Pathfinder.jl itself rather than their dependencies. Would that be enough of a speed-up to depend on Pathfinder itself?

```julia
julia> @time using BAT
  7.763856 seconds (27.45 M allocations: 1.726 GiB, 5.12% gc time, 50.26% compilation time)

julia> @time using AbstractDifferentiation
  0.401405 seconds (1.53 M allocations: 84.648 MiB, 2.92% gc time, 92.30% compilation time)

julia> @time using GalacticOptim
  0.378646 seconds (1.45 M allocations: 120.713 MiB, 18.09% compilation time)

julia> @time using Folds
  0.186595 seconds (692.59 k allocations: 38.897 MiB, 11.13% gc time, 50.81% compilation time)

julia> @time using Pathfinder
  0.193904 seconds (601.56 k allocations: 40.755 MiB, 16.95% compilation time)
```

oschulz:
> but if this is a feature you urgently need, then yes, this is in general fine.

Yes, we have some current physics analysis use cases that would profit a lot from faster MCMC init.

> @time using BAT [...]  7.783539 s

Yeah :-( I hope that'll go down, we currently have some expensive hard dependencies like Zygote that I want get rid of (as mandatory deps).

>  if we make these 4 dependencies optional for Pathfinder

Yes, GalacticOptim (because of SciMLBase), AbstractDifferentiation, PSIS and Folds seem indeed to be the main "cost factors".

We may actually not need PSIS in BAT here directly - that is, we'll probably offer it as an algorithm for importance sampling in general, but I'll try to separate that from VI-posterior approximation (possibly by a package boundary in the future).

If it's Ok with you, I'd merge this for now so we can try it out in some immediate use cases (I've set you as author for the commit). Obviously code duplication is bad, so I'd remove it from BAT as soon as we can use it from Pathfinder and PDMats without load-time impact (PDMats we need anyway, of course).

oschulz:
> I do think it needs more thorough attribution consistent with the MIT license and ideally with a URL linking to the Pathfinder commit the code was taken from

I've added

```
# Code imported from Pathfinder.jl (https://github.com/sethaxen/Pathfinder.jl)
# under MIT License. Original author is Seth Axen <seth.axen@gmail.com>.
```

I hope that's fine (since BAT.jl is under the same licence). And I've added a ToDo-comment so we won't forget to remove it later on.

src/statistics/woodbury_pdmat.jl
```diff
@@ -0,0 +1,173 @@
+# Code imported from Pathfinder.jl (https://github.com/sethaxen/Pathfinder.jl)
+# under MIT License. Original author is Seth Axen <seth.axen@gmail.com>.
```
sethaxen, line 2:
```suggestion
# under MIT License.
# Copyright (c) 2021 Seth Axen <seth.axen@gmail.com> and contributors
```

src/statistics/lbfgs_inverse_hessian.jl
```diff
@@ -0,0 +1,93 @@
+# Code imported from Pathfinder.jl (https://github.com/sethaxen/Pathfinder.jl)
+# under MIT License. Original author is Seth Axen <seth.axen@gmail.com>.
```
sethaxen, line 2:
```suggestion
# under MIT License.
Copyright (c) 2021 Seth Axen <seth.axen@gmail.com> and contributors
```

sethaxen (changes requested):

sethaxen:
Technically the copyright notice should be included explicitly, so I've added that.

sethaxen:
> We may actually not need PSIS in BAT here directly - that is, we'll probably offer it as an algorithm for importance sampling in general, but I'll try to separate that from VI-posterior approximation (possibly by a package boundary in the future).

The 2 main reasons to use PSIS are 1) to better draw samples from a variational approximation to the posterior 2) to diagnose the quality of the approximation. It's necessary also for multi-path Pathfinder to be a major improvement over single-path (and multi-path is highly recommended, even for MCMC initialization). See http://proceedings.mlr.press/v80/yao18a.html I think we can make PSIS a little more lightweight so that it wouldn't affect your load times at all, even after refactoring.

oschulz:
> The 2 main reasons to use PSIS are 1) to better draw samples from a variational approximation to the posterior 2) to diagnose the quality of the approximation.

Definitely! We also do have a simple prior-based importance sampler in BAT (for easy cases with very informative priors), I think PSIS could be useful there as well. This may live in a sub-package of BAT in the future though - BAT has really grown way to big, so it needs to become more modular like the Turing packages. Optimizing dependencies will be the tricky part. :-)

oschulz:
> Technically the copyright notice should be included explicitly, so I've added that.

Thanks!

Ok to merge like this from your side @sethaxen ?

oschulz:
> I think we can make PSIS a little more lightweight so that it wouldn't affect your load times at all, even after refactoring.

That would be awesome - other statistics packages could likely profit from the method as well, after all.

src/statistics/lbfgs_inverse_hessian.jl
```diff
@@ -0,0 +1,94 @@
+# Code imported from Pathfinder.jl (https://github.com/sethaxen/Pathfinder.jl)
+# under MIT License.
+Copyright (c) 2021 Seth Axen <seth.axen@gmail.com> and contributors
```
sethaxen, line 3:
```suggestion
# Copyright (c) 2021 Seth Axen <seth.axen@gmail.com> and contributors
```

sethaxen (approved):
Sorry, forgot to add a pound sign. All good on my end.

oschulz:
Thanks a lot @sethaxen !

oschulz:
> BAT has really grown way to big, so it needs to become more modular like the Turing packages. Optimizing dependencies will be the tricky part.

Memo for BAT modularisation plans: The load time of the BAT code itself is actually not bad, it's all about the dependencies, so we'll mainly have to modularize based on the deps of BATs functional domains:

```julia
julia> @time using AdaptiveRejectionSampling, AdvancedHMC, ArgCheck, CPUTime, ChainRulesCore, ChangesOfVariables, Clustering, Colors, DataStructures, DensityInterface, DiffResults, Distributions, DistributionsAD, DocStringExtensions, DoubleFloats, FFTW, FillArrays, FiniteDiff, ForwardDiff, HypothesisTests, IntervalSets, InverseFunctions, KernelDensity, LaTeXStrings, Measurements, NLSolversBase, NamedArrays, Optim, PDMats, ParallelProcessingTools, Parameters, PositiveFactorizations, PrettyTables, ProgressMeter, QuadGK, Random123, RecipesBase, Requires, Sobol, SpecialFunctions, StableRNGs, StaticArrays, StatsBase, StructArrays, Tables, TypedTables, Zygote, ZygoteRules, ElasticArrays, ArraysOfArrays, EmpiricalDistributions, ValueShapes, ForwardDiffPullbacks

 10.788600 seconds (22.21 M allocations: 1.367 GiB, 5.34% gc time, 57.44% compilation time)

julia> @time_imports using BAT
    124.5 ms  BAT
```

PR #356, opened 2022-05-17
Use `track_order = true` in IO via HDF5.jl

### LINKED ISSUE #355: Order of Samples not preserved in IO via HDF5.jl
The example from the test set "hdf5" fails, if the order of the parameters in the prior is not alphabetically sorted.

## MWE

I just changed `a` and `b` in `example_posterior`:
```julia


using Test
using BAT
using StableRNGs
using Distributions
using ValueShapes

import HDF5


function BAT.example_posterior()
    rng = StableRNGs.StableRNG(0x4cf83495c736cac2)
    prior = NamedTupleDist(
        b = [4.2, 3.3], # switched with `a`
        a = Exponential(),  
        c = Normal(1, 3),
        d = [Weibull(), Weibull()],
        e = Beta(),
        f = MvNormal([0.3, -2.9], [1.7 0.5; 0.5 2.3])
    )
    n = totalndof(varshape(prior))
    A = randn(rng, n, n)
    likelihood = varshape(prior)(MvNormal(A * A'))
    PosteriorDensity(likelihood, prior)
end

mktempdir() do tmp_datadir
    results_filename = joinpath(tmp_datadir, "results.hdf5")
    samples = bat_sample(BAT.example_posterior(), MCMCSampling(nsteps = 1000, strict = false)).result
    bat_write(results_filename, samples)
    samples2 = bat_read(results_filename).result
    @test samples == samples2
end
```

lmh91:
This fixes #355

The read-in requires some modification of internal HDF5 parameters at the moment.
I also opened an issue in HDF5.jl: https://github.com/JuliaIO/HDF5.jl/issues/939

Please have a look @oschulz

### DIFF
--- a/src/io/bat_io.jl
+++ b/src/io/bat_io.jl
@@ -38,7 +38,21 @@ function bat_read(src::Union{AbstractString,Tuple{AbstractString, AbstractString
     ftype = _file_type_from_extension(src)
     if ftype == "HDF5"
         _h5io_open(src, "r") do input
-            (result = _h5io_read(input),)
+            #=
+                Currently (HDF5.jl - v0.16.9), the keyword `track_order` is ignored in read-in. 
+                Thus, HDF5.IDX_TYPE[] has to be set manually.
+                The try-catch-block is necessary in order to be able to load old files.
+            =#
+            prev = HDF5.IDX_TYPE[] 
+            HDF5.IDX_TYPE[] = HDF5.API.H5_INDEX_CRT_ORDER
+            r = try
+                (result = _h5io_read(input),)
+            catch err
+                HDF5.IDX_TYPE[] = prev
+                (result = _h5io_read(input),)
+            end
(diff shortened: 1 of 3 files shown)


oschulz:
> The read-in requires some modification of internal HDF5 parameters at the moment.

Will this fail with older versions of HDF5.jl? Can we check if support for `track_order = true` is present?

lmh91:
> Will this fail with older versions of HDF5.jl? Can we check if support for `track_order = true` is present?

Ah, right... I guess it will fail with versions of HDF5 <= v0.16.2. `track_order` was introduced in `v0.16.3`.

This 
```julia
isdefined(HDF5, :class_getproperty) && isdefined(HDF5, :FileCreateProperties) && 
    in(:track_order, HDF5.class_propertynames(HDF5.FileCreateProperties))
```
might do it.

oschulz:
Looks good to me, thanks!

PR #357, opened 2022-05-24
add vsel processing and bat_marginalize

Micki-D:
I implemented functions to marginalize DensitySampleVectors
the usage is e.g:
>>julia  marg_samples = bat_marginalize(samples, :(a[1,2,3]))

where 'samples' are the shaped or unshaped samples to be marginalized and the second argument is the desired columns that are cut out of the original data. 
The logd of the new samples is set to NaN, the info and aux fields of the original data is copied. 

the tests are a work in progress.

### DIFF
--- a/src/BAT.jl
+++ b/src/BAT.jl
@@ -33,6 +33,7 @@ using IntervalSets
 using InverseFunctions
 using KernelDensity
 using LaTeXStrings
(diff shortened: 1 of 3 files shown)


Micki-D:
do the tests error because of something I did?

PR #364, opened 2022-06-21
WIP: changed to transform_and_unshape in bat_sample_impl

Cornelius-G:
Line 41 in `chain_pool_init.jl` didn't work for `trafo = NoDensityTransform`.

By changing from `bat_transform` to `transform_and_unshape` in `mcmc_sample.jl` this is solved.
However, in `mcmc_sample.jl` I now completely replaced the old trafo by the unshaping trafo. This gives the correct results with the only difference being that now also for `trafo = NoDensityTransform` the `samples.result_trafo` are unshaped (as they are for `PriorToGaussian` and `PriorToUniform`). Before, for `NoDensityTransform` `samples.result` and `samples.result_trafo` were the same. 

I find this now a bit more consistent since users can now be sure that `samples.result_trafo` is always unshaped. 
But this would require some adaptions to the other sampling algorithms to behave the same way (shouldn't be that much).
I'm not sure if we want to go in that direction or if I should just pass the "unshaping" trafo only to the `initval` algorithm.

### DIFF
--- a/src/samplers/mcmc/chain_pool_init.jl
+++ b/src/samplers/mcmc/chain_pool_init.jl
@@ -22,11 +22,14 @@ end
 
 export MCMCChainPoolInit
 
-apply_trafo_to_init(trafo::Function, initalg::MCMCChainPoolInit) = MCMCChainPoolInit(
+
+function apply_trafo_to_init(trafo::Function, initalg::MCMCChainPoolInit)
+    MCMCChainPoolInit(
     initalg.init_tries_per_chain,
     initalg.nsteps_init,
     apply_trafo_to_init(trafo, initalg.initval_alg)
-)
+    )
+end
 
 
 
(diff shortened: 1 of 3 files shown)


oschulz:
> now also for trafo = NoDensityTransform the samples.result_trafo are unshaped (as they are for PriorToGaussian and PriorToUniform). 

That's good, we want that anyway - it's the space the sampler run in, after all.

oschulz:
> I find this now a bit more consistent since users can now be sure that samples.result_trafo is always unshaped. [...] I'm not sure if we want to go in that direction

Yes, definitely!

oschulz:
This replaces #360, right?

Cornelius-G:
> This replaces #360, right?

Yep. But I'm currently still testing a bit more with your latest commits. Not yet sure if my proposed changes here have some unexpected effects.

oschulz:
Closed?

oschulz:
Was this superseded by #369?

PR #369, opened 2022-06-24
unshape result_trafo and add initval_alg to MCMCChainPoolInit

Cornelius-G:
As discussed in #364 all samplers now return unshaped samples in the field `result_trafo`, also for `DoNotTransform()` 
This one should replace #364 (sorry for opening another PR, had to make a new branch)

@oschulz, there is something strange going on with the trafos in "mcmc_sample.jl" line 91 following.
If `trafo` is something like `identity ∘ identity` broadcasting with `inverse(trafo).(samples_trafo)` fails and is not needed since `inverse(trafo)(samples_trafo)` works. But for the other trafos the broadcasting is needed (see [stacktrace.txt](https://github.com/bat/BAT.jl/files/8974239/stacktrace.txt)).

Strangely, in "importance_sampler.jl" and "ellipsoidal_nested_sampling.jl" the same code is used but works with broadcasting for all trafos. It seems to me there is an issue if `DensitySampleVector` contains something in `info`. At least, that's the only difference in the samples for mcmc and the importance sampler. Do you have any idea what might be going on there?


oschulz:
> Do you have any idea what might be going on there?

No, that's weird. Can you try to make a minimal code snippet to reproduce?

oschulz:
This should fix it, I hope: #370

oschulz:
> However, for using this to start sampling in/very close to the mode ...  tuner scale ...

We need to make tuner scale adaption more dynamic anyway.

oschulz:
Thanks!

PR #372, opened 2022-06-28
Add bat_report, refactor SampledMeasure

oschulz:


oschulz:
Allows for faster creation of `SampledMeasure` (which will have more use cases beyondd reporting). The new `bat_report` provides a basis for configurable reporting in the future (problematic with the current `Base.show(::SampledMeasure)` approach since `show` can't take arguments like parameter selection, etc.).

PR #380, opened 2022-07-28
Add flat index support for bat_marginalize()

Micki-D:
Adds support for doing 

marg_s = BAT.bat_marginalize(some_shaped_samples, i)

where `i` is an Integer, Vector or Tuple of Integers or Unit Range.

The new marginalized samples will have `N` ScalarShaped parameters with the keys being encoded according to the keys of the input data. `N` being the number of columns selected in the marginalization process.

If a range is input, Expressions corresponding to the selected columns will be generated up to a range of 5. For larger ranges strings will be used to create the encoded symbols to save memory.

### DIFF
--- a/examples/dev-internal/plotting_examples.jl
+++ b/examples/dev-internal/plotting_examples.jl
@@ -3,6 +3,7 @@
 using BAT
 using Distributions
 using IntervalSets
+using DensityInterface
 
 # ## Generate samples to be plotted
 likelihood = logfuncdensity(params -> begin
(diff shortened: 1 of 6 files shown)


oschulz:
@Cornelius-G can you give this a test?

oschulz:
@Cornelius-G, let me know when you're done checking this out?

Cornelius-G:
@oschulz I did now check this with our examples and the flat indexing seems to work again! 
So from my side, this would be ready to be merged. Thanks @Micki-D

oschulz:
Thanks @Micki-D and @Cornelius-G !

PR #383, opened 2022-08-08
Add bat_transform support for samples

oschulz:
@cescalara


cescalara:
Nice, thanks! Looking forward to checking this out

oschulz:
Merging this so it can be used immediately, test failures seem to be due to some weird Conda problems that don't occur locally. Will figure this out in a separate PR.

PR #384, opened 2022-08-31
Update to HDF5 release v0.16.11

lmh91:
Adapt to the breaking changes in the lastest HDF5 release.

Note that I modified the `example_posterior` to make the tests actually sensitive to the order.


oschulz:
Thanks a lot!

PR #386, opened 2022-09-04
Switch Recipes PriorOverview _all_exprs -> _all_active_exprs

waldie11:
A element of prior can be const. So far these are not plotted (and there probably is no need to anyways). This patch reenables to plot overview on priors which have constant elements aka number of elements in totalndof/_all_active_exprs is smaller than in implementation _all_exprs.

One could open up a TODO to plot prior overviews by Symbol and Expr as well (those are the accepted vsel arguments for a single plot)

Additional work needs to be done if a element should be displayed with the same shape, as it is provided by a ShapedAsNT prior.

### DIFF
--- a/src/plotting/recipes_prior_overview.jl
+++ b/src/plotting/recipes_prior_overview.jl
@@ -15,8 +15,8 @@
         vsel = reduce(vcat, vsel)
     end
     vsel = vsel[vsel .<=  totalndof(BAT.varshape(prior))]
-    all_exprs = _all_exprs(prior)
-    vsel = all_exprs[vsel]
+    all_active_exprs = _all_active_exprs(prior)
+    vsel = all_active_exprs[vsel]
 
     xlabel = string.(vsel)
     ylabel = ["p($l)" for l in xlabel]
--- a/src/plotting/recipes_prior_overview.jl
+++ b/src/plotting/recipes_prior_overview.jl
@@ -2,7 +2,7 @@
 
 @recipe function f(
     prior::NamedTupleDist;
-    vsel=collect(1:5),
+    vsel::AbstractVector=collect(1:5),
     bins = 200,
     diagonal = Dict(),
     upper = Dict(),
--- a/src/plotting/recipes_prior_overview.jl
+++ b/src/plotting/recipes_prior_overview.jl
@@ -109,20 +109,22 @@
 
 end
 
-function _all_exprs(dist::NamedTupleDist)
+function _all_active_exprs(dist::NamedTupleDist)
     vs = varshape(dist)
     accs = vs._accessors
     syms = keys(accs)
-    lengths = length.(values(accs))
     vsel = Any[]
 
     for (i,sym) in enumerate(syms)
         vsel_tmp = Any[]
 
-        if lengths[i] == 1 
+	lengths_i = getproperty(vs, sym).len
+	if lengths_i == 0
+           skip
+        elseif lengths_i == 1 
             push!(vsel_tmp, Meta.parse("$sym"))
         else
-            for id in 1:lengths[i]
+            for id in 1:lengths_i
                 push!(vsel_tmp, Meta.parse("$sym[$id]"))
             end
         end


oschulz:
@Cornelius-G , could you take a look at this? @Micki-D , you too, maybe?

waldie11:
Essentially, if you want to test this you need a DensitySampleVector `samples`, based on a prior with const values.
I.e.
```julia
using BAT
using ValueShapes
using DensityInterface
using Distributions
using Plots

N = 100

prior = NamedTupleDist(ShapedAsNT, #<- optional, but required for unshaped.(samples)
    a = 1,
    b = reshape(product_distribution(fill(Normal(1),6)),2,3),
    c = Uniform(0,1),
)
vs = valshape(prior)

X = rand(prior,N)
usamples = DensitySampleVector(unshaped.(X),logdensityof(prior).(X))
samples = vs.(usamples)

plot(usamples, 1)
plot(samples, :(b[1]))
plot(samples, vsel=1:3)
```

The suggestions concern functionality not available atm
```julia
plot(samples, vsel=[:(b[1]),:(b[2])]
plot(samples, :(b[1,2])) #<- consider b is a 2x3 Matrix
```
Plotting by index is still unsupported
```julia
plot(samples,1)
```

Micki-D:
Could you elaborate on your last suggestion about the element being displayed in the same shape as it is provided in the pior? I don't quite understand what you mean. 

It is possible to select certain parameters from a shaped or unshaped `DensitySampleVector` via `bat_marginalize()` and have them stored in a new  `DensitySampleVector` in the same shape.  The documentation for this is in lines 225-279 of https://github.com/bat/BAT.jl/pull/380/files

waldie11:
> , I believe we'd have to create an array like fill(const_value, N) or so at some point,

I wouldn't recommend following exactly that `fill` approach (although there is neat FillArrays.jl already used somewhere). Thing is, if a ConstValueShape on `a` holds non-scalar values, there is no guarantee all are equal.
```julia
prior = NamedTupleDist(ShapedAsNT,
    a = reshape(1:6,2,3),
    ....
)
```

> so I hope it isn't very relevant in practice.
> So far these [const valued elements of prior] are not plotted (and there probably is no need to anyways). 

Coming to think of it, I think in times it might be neat to show correlation of a const valued to a non-const - so essentially a horizontal / vertical line with contour.

> If you wish to select a const value, you can do so via an expression like :(a[1,2]) if a is a ConstValueShape holding a 2D Matrix for example.

This is not possible atm in PriorOverview on non-const nor const whilst it is possible to visualize single elements of prior thatway. I think at least
[this one](https://github.com/bat/BAT.jl/blob/main/src/plotting/recipes_prior_overview.jl#L17) is to blame.

The latter is not currently supported, right?
```julia
plot(samples, :(b[1,2]))
```
> ERROR: ArgumentError: Symbol :b[1, 2] found not in NamedTupleShape(ShapedAsNT,

waldie11:
> Could you elaborate on your last suggestion about the element being displayed in the same shape as it is provided in the pior? I don't quite understand what you mean.

In my mind it would be very tempting to see the output of either
```julia
plot(samples, :(b[2,1]))
plot(samples, 2)
```
firstly to work of course, but also in the way that they label axis in this way by default
```julia
plot(usamples,2,xlabel="b[2,1]",ylabel="p(b[2,1])")
```

B21

Featuring sth like
```julia
plot(samples, [vsel=]:(b[:,1]) )
```
would be marvelous, but I think
```julia
plot(samples, vsel=[eval(Meta.parse(":(b[$i,1])")) for i in 1:2])
```
is enough from a user perspective.

Unfortunately I don't have resources to approach this atm.

waldie11:
@oschulz
Would `Base.collect(vs::ConstValueShape) = vs.shape.value` be sth ValueShapes.jl could provide?

oschulz:
> Would Base.collect(vs::ConstValueShape) = vs.shape.value be sth ValueShapes.jl could provide?

Hm, no, I don't think that is a good fit, semantically.

Micki-D:
> I wouldn't recommend following exactly that `fill` approach (although there is neat FillArrays.jl already used somewhere). Thing is, if a ConstValueShape on `a` holds non-scalar values, there is no guarantee all are equal.

I agree, but somehow the values to be plotted need to be put into a shape that can be plotted, and in my understanding that always needs to be as x-y pairs of values at some point. 

> The latter is not currently supported, right?
> 
> ```julia
> plot(samples, :(b[1,2]))
> ```
> 
> > ERROR: ArgumentError: Symbol :b[1, 2] found not in NamedTupleShape(ShapedAsNT,

If `b` in this example is a `ConstValueShape`, unfortunately this isn't currently supported, correct.

I will reach out to @Cornelius-G and discuss the issues you raised and your suggestion about the plotting of const values with with him.

Cornelius-G:
Hi, I just tested the examples from @waldie11 in https://github.com/bat/BAT.jl/pull/386#issuecomment-1236379803. With PR #380 now everything works as expected, I guess. Only the line `plot(samples, :(b[1,2])) #<- consider b is a 2x3 Matrix` is not yet supported. Currently, only the linear indexing of multidimensional parameters like `b` is possible, i.e. `plot(samples, :(b[3]))` since `b[1,2] = b[3]` in this example. 

@Micki-D could you give it a try to implement also the cartesian indexing of multidimensional arrays in plotting?

Micki-D:
> @Micki-D could you give it a try to implement also the cartesian indexing of multidimensional arrays in plotting?

Will do!

oschulz:
Let's bring this home. :-) @Micki-D and @Cornelius-G , can you check if this is good to merge?

Micki-D:
Looks good from my side! 
I'm still working on the cartesian indexing for the vsel argument

oschulz:
> I'm still working on the cartesian indexing for the vsel argument

So should we hold off on this, or will that be a separate PR?

Micki-D:
Oh, no that will be a separate PR!
In my opinion we can merge this PR here.

oschulz:
Ok, here we go ...

oschulz:
Replaced by #392 to resolve merge conflicts.

PR #392, opened 2022-11-20
Replace _all_exprs by _all_active_exprs in prior overview

oschulz:
Replaces #386


oschulz:
@Micki-D, this resolves the merge conflicts in #386, but now we have both `_all_exprs` and `_all_active_exprs` - do we need both? Can you take a look?

oschulz:
@Micki-D I just rebased this - can you take a look at it (see above)?

Micki-D:
Crap, sorry I didn't see the notification, I'll check this out asap!

oschulz:
Thanks!

Micki-D:
Hi, this looks good to me, we just need to rename the function in 'MarginalDist_utils.jl', [line 211](https://github.com/bat/BAT.jl/blob/ae6edf6e19cfe887855173b788a3226cfd663a28/src/plotting/MarginalDist_utils.jl#L211) '_all_exprs()' to '_all_active_exprs()'. I ran the example plots with that change and it runs fine.
Should I open another PR with that change?

PR #409, opened 2023-06-30
WIP MCMC with transforms Interface

waldie11:
FYI
@oschulz @Cornelius-G


oschulz:
Can you check the

```
Base.Meta.ParseError("parsing error in src/samplers/transformed_mcmc/struct_list.jl:34: missing comma or } in argument list")
```

CI error on Julia 1?

src/samplers/transformed_mcmc/example.jl
```diff
@@ -0,0 +1,59 @@
+using BAT
```
oschulz, line 1: Examples should live in "examples", not in "src" :-)

src/samplers/transformed_mcmc/mcmc_algorithm.jl
```diff
@@ -0,0 +1,244 @@
...
+    abstract type TransformedMCMCBurninAlgorithm
+
+Abstract type for MCMC burn-in algorithms.
+"""
+abstract type TransformedMCMCBurninAlgorithm end
+export TransformedMCMCBurninAlgorithm
+
+
+
+@with_kw struct TransformedMCMCIteratorInfo
```
oschulz, line 41: Stuff like this we don't need to duplicate.

src/samplers/transformed_mcmc/mcmc_algorithm.jl
```diff
@@ -0,0 +1,244 @@
+# This file is a part of BAT.jl, licensed under the MIT License (MIT).
+
+
+
+function get_mcmc_tuning end #TODO: still needed
+
+
+"""
+    abstract type TransformedMCMCInitAlgorithm
```
oschulz, line 9: We don't need to duplicate abstract types like these, I think.

oschulz (changes requested):

oschulz:
I've updated this to use the new BATContext. "transformed_example.jl" doesn't fully run through, but I'm not sure if that's related.

oschulz:
`TransformedAdaptiveMHTuning` in the example works now, but

```
r_hmc = @time BAT.bat_sample_impl(posterior, MCMCSampling(mcalg=HamiltonianMC(), nchains=4, nsteps=4*20000), context)
```

in the example doesn't work yet, but that's to be expected currently, right?

I think `BATContext`-wise this is fine now.

Cornelius-G:
> `TransformedAdaptiveMHTuning` in the example works now, but
> 
> ```
> r_hmc = @time BAT.bat_sample_impl(posterior, MCMCSampling(mcalg=HamiltonianMC(), nchains=4, nsteps=4*20000), context)
> ```
> 
> in the example doesn't work yet, but that's to be expected currently, right?

Hm, for me this line works. As it should, because that is just the "good old" HMC smapling, not using the new transformations.

oschulz:
Ah - well, in any case, I think the adaption to BATContext is complete.

src/samplers/transformed_mcmc/mcmc_tuning/mcmc_proposalcov_tuner.jl
```diff
@@ -58,7 +58,7 @@ end
 
 # this function is called once after each tuning cycle
 g_state = nothing
-function tuning_update!(tuner::TransformedProposalCovTuner, chain::MCMCIterator, samples::DensitySampleVector)
+function tuning_update!(tuner::TransformedProposalCovTuner, chain::TransformedMCMCIterator, samples::DensitySampleVector)
```
Cornelius-G, line 61: I guess we need to revert this change, because the CI tests fail due to this line.
`TransformedMCMCIterator` is a subtype of `MCMCIterator`. Dispatch should be fine just using the `tuner` argument.

oschulz: Sure, we can revert it. But why does it cause errors? For a `TransformedProposalCovTuner` the chain must be a `TransformedMCMCIterator`, right?

oschulz:
@Micki-D needs to be updated to API changes.

oschulz:
Closing in favor of #441

PR #411, opened 2023-07-06
Use AutoDiffOperators and add BATContext

oschulz:
Pass `BATContext` objects around as the optional last argument of `bat_...` functions, instead of `rng` as the optional first argument. The context contains both an RNG and AD selection.


oschulz:
#409 should be fairly easy to adapt, mainly be changing `bat_..._impl(rng::AbstractRNG, ...)` to `bat_..._impl(..., contect::BATContext)`.

`get_rng(context)` will return an `AbstractRNG` where needed.

PR #415, opened 2023-07-10
Fix HierarchicalDistributions w/ dof smaller than length

waldie11:
For some distributions length and dof do not coincide. This enables usage of e.g. Dirichlet distributions in HierarchicalDistributions


oschulz:
Thanks!

waldie11:
With pleasure :wink:

PR #416, opened 2023-07-11
Fix revert to functional rand(::HierarchicalDistribution)

waldie11:
Only checked the transform in #415 , so now reenable functionality to draw random sample from HierarchicalDistributions / anything calling _hd_split with eff_totalndof ≠ length in primary_dist.
No guarantees that there are some calls remaining which should rather work on newly introduced _hd_split_efftotalndof

```
using BAT
using ValueShapes
using Distributions
using InverseFunctions

d = HierarchicalDistribution(NamedTupleDist(
       a=truncated(Logistic(),0,Inf),
)) do v
       NamedTupleDist(
             b=Dirichlet(10,1/v.a),
       )
end

trafo = BAT.DistributionTransform(Normal,d)
inverse(trafo)(rand(Normal(), BAT.eff_totalndof(d))

rand(d)
rand(unshaped(d))
trafo(rand(d))

```

### DIFF
--- a/src/distributions/hierarchical_distribution.jl
+++ b/src/distributions/hierarchical_distribution.jl
@@ -189,14 +189,25 @@ end
 =#
 
 function _hd_split(ud::UnshapedHDist, x::AbstractVector{<:Real})
-    @argcheck eff_totalndof(ud) == length(eachindex(x))
+    @argcheck length(ud) == length(eachindex(x))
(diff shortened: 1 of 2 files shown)


oschulz:
Ok, I'll revert https://github.com/bat/BAT.jl/pull/415 until we have this consistent. :-)

PR #417, opened 2023-07-11
Fix HierarchicalDistributions w/ dof smaller than length

waldie11:


waldie11:
Only checked the transform in https://github.com/bat/BAT.jl/pull/415 , so now reenable functionality to draw random sample from HierarchicalDistributions / anything calling _hd_split with eff_totalndof ≠ length in primary_dist.
No guarantees that there are some calls remaining which should rather work on newly introduced _hd_split_efftotalndof

```
using BAT
using ValueShapes
using Distributions
using InverseFunctions

d = HierarchicalDistribution(NamedTupleDist(
       a=truncated(Logistic(),0,Inf),
)) do v
       NamedTupleDist(
             b=Dirichlet(10,1/v.a),
       )
end

trafo = BAT.DistributionTransform(Normal,d)
inverse(trafo)(rand(Normal(), BAT.eff_totalndof(d)))

rand(d)
rand(unshaped(d))
trafo(rand(d))
```

waldie11:
@oschulz sry for confusing with #416 - only aborted that one as I wanted to have the cleaner commit history possible at that time.  the "no guarantees" is also more conservative than necessary, I think :wink:

Iam rather convinced that this should do the job.

oschulz:
Thanks again! :-)

Issue #426, opened 2023-09-27
`HamiltonianMC()` is failed on the tutorial model


takuizum:
Hello, everyone,

I'm trying out BAT.jl out of curiosity.
I found that `HamiltonianMC()` does not work on the tutorial model.
More specifically, in the tutorial, the following syntax is used to sample from the posterior of the model.
```
samples = bat_sample(posterior, MCMCSampling(mcalg = MetropolisHastings(), nsteps = 10^5, nchains = 4)).result
```

Changing `mcalg = HamiltonianMC()`, return an error.
```
ERROR: Density evaluation with logdensityof failed at [NaN, NaN, NaN, NaN, NaN], must not evaluate to NaN, density is DistMeasure(objectid = 0x81ff2597fbd0fd5d, varshape = ValueShapes.ArrayShape{Real, 1}((5,)))
```

It would be greatly appreciated if someone could help with this problem.

Thank you,

oschulz:
Hi, sorry @takuizum , your issue got overlooked somehow. I'll check what's wrong there.

oschulz:
This seems to be an intermittent error, I'm tracking it down.

oschulz:
This seems to occur when the parameters go to regions in which the expected counts in one of the histograms bins becomes zero. In that case, this happens:

```julia
julia> f = v -> logpdf(Poisson(v), 1)
#32 (generic function with 1 method)

julia> f(0.0)
-Inf

julia> ForwardDiff.derivative(f, 0.0)
NaN
```

We didn't see this in the past, I wonder if something has changed in the implementation of `Distributions.Poisson`.

oschulz:
@takuizum this should be fixed now in  the tutorial.

PR #427, opened 2023-10-20
Fix numerical instability in tutorial likelihood

oschulz:
Expected counts of zero result in NaN log-likelihood gradient.


oschulz:
Fixes #426

PR #439, opened 2024-05-02
Update 'Allow Optim.Options to pass to BAT.find_mode'

Micki-D:
This updates PR #423 to the current main branch.


Micki-D:
I created a new PR, since I thought this would be cleaner than changing PR #423. Let me know if I should do it the other way.

oschulz:
> I created a new PR,

That's fine!

Micki-D:
I am confused about the Julia 1.6-ubuntu-latest tests failing. 
In the compilation of the environment, it gives a warning that the exact package versions could not be loaded, and indeed, when comparing the compiled package versions to the other test runs that pass, the failing one has different versions. 
Most are lower versions, but there are some higher versions than in the passing test. 

Could there be some issue with the job setup?

oschulz:
@Micki-D can you try to run tests manually on v1.6

oschulz:
Why did you close this one @Micki-D ?

Micki-D:
Ah beans. I did not mean to do so. I'll reopen this @oschulz .

PR #441, opened 2024-06-03
Rename transformed (Update for Measure terminology)

Micki-D:
Updates the original PR #409 with the new measure terminology.


Micki-D:
So the failing tests are due to errors in the generation of viable MCMC chains. 
But I am unsure if this is due to issues with the Measure terminology or due to issues in the original PR #409.
Were there any known issues with the transformed MCMC algorithms @Cornelius-G ?

Micki-D:
Also, is there a convention for when to use 'X_density' or 'X_measure' or 'm' as variable names?  
It's unclear to me from the existing code.

PR #462, opened 2024-11-15
Include sample_z in _cleanup_samples()

Micki-D:
This fixes a bug that arises when '_cleanup_samples()' is called when the last proposed sample in the chain has been accepted. 
This would cause the samples stored in the chain_state to be cleaned up, but the sample_z would remain in the old state, causing faulty subsequent steps.


oschulz:
Thanks!

PR #467, opened 2024-12-16
Refactor z position update in transformed MCMC

Micki-D:
This PR changes the way the `sample_z`object is updated during transformed mcmc. 
Now, a new `DensitySampleVector` with the updated z samples is created, which is then used in a new mcmc chain state via `@set` from `Accessors.jl`


oschulz:
Thanks!

PR #468, opened 2025-02-12
Hmc proposed sample handling refactor

Micki-D:
This adds a modified version of the function `transition()` from AdvancedHMC.jl to make the points that are proposed during the HMC algorithm accessible in BAT. 

Also refactors `StanHMCTuning()` to work in the transformation based sampling framework. 

Original Function: 
https://github.com/TuringLang/AdvancedHMC.jl/blob/878efce497852dfe0b6918e9de4b825706fa848a/src/trajectory.jl#L674


oschulz:
Can you remote the stuff that only changes whitespace from the PR?

oschulz:
Looks good to me, just have to figure out the test failures.

PR #470, opened 2025-03-23
Adjust for API changes in AdvancedHMC.jl 

Micki-D:
This checks if `rand_momentum()` from AdvancedHMC.jl is defined and if true uses this over the deprecated `rand()` method in older versions (<0.7) of AdvancedHMC.jl.


oschulz:
Test failures were due to outdated dependency bounds of MGVI.jl - that's fixed now, let's see if tests pass now.

PR #473, opened 2025-03-27
Enable ensemble based sampling for MCMC algorithms

Micki-D:
This modifies the infrastructure for MCMC sampling to handle ensemble based Markov Chains. 
This means that each 'MCMCChainState' can now have multiple walkers each with a proposed and current point. 

Also the handling of these current and proposed points is updated:
Now 'MCMCChainState' has the fields 'current' and 'proposed' which are named tuples with the fields 'x' and 'z' that each hold a 'DensitySampleVector' containing the samples in the X or Z space according to the space transformation of the Chain. The i-th entries in these 'DensitySampleVectors' hold the samples corresponding to the i-th walker of the chain. 

Also new to 'MCMCChainState' are the fields 'accepted' which is a Vector of Booleans, where the i-th entry states whether the i-th walker of the chain accepted its proposed point in the current step. 
And lastly the field 'output' which is a 'DensitySampleVector' where the i-th entry holds the sample produced by the i-th walker in the current step. Either an accepted point or the rejected point.


Micki-D:
The sampling runs with the new infrastructure, but I have not yet adjusted the tests. 
The sampling now returns a Vector of `DensitySampleVectors` (`DSV`) with the i-th `DSV` holding the samples produced by the i-th walker. 
But currently I mix the samples back together in the output of `bat_sample()` to adhere to the existing API. 

I think we should keep it this way, but add a new field to the `result` NamedTuple that is returned by `bat_sample()` with this new field holding the sorted `DSV`'s of the walkers .

oschulz:
Why is Documenter unhappy?

PR #483, opened 2025-07-29
Exclude plotting code from codecov

fhagemann:
The code coverage might be underestimating how many lines of code are actually covered, because many of them are defining plot recipes.

This PR excludes these files from the `codecov` tests to provide a more realistic number on the code coverage.


oschulz:
Nice, thanks!

PR #477, opened 2025-05-27
Adding plotting functions for Makie

philippeller:
# Overview

This PR is adding plotting functions to BAT for use with Makie.jl

Some basic plot types are implemented for 1d and 2d. The plotting is fully extensible by providing structs of abstract type `plot_cfg_1d ` and `plot_cfg_2d`, respectively, together with corresponding function implementations:

```
function plot1d!(ax::Axis, cfg::plot_cfg_1d, x::AbstractArray, w::Union{Real, AbstractArray}=1)
    scatter!(ax, x, y, markersize=sqrt.(w./mean(w)), color=cfg.color)
end
```
or, respectively, 
```
function plot2d!(ax::Axis, cfg::plot_cfg_2d, x::AbstractArray, y::AbstractArray, w::Union{Real, AbstractArray}=1)
    scatter!(ax, x, y, markersize=sqrt.(w./mean(w)), color=cfg.color)
end
```

(And really should be extended in the future, of course.)

# Example usage:

```julia
using BAT
using MeasureBase
using DensityInterface
using Distributions
using LinearAlgebra

params = (a=1, b=1, c=1)
priors = (a=Normal(), b=Uniform(-2,2), c=Cauchy())

function fwd(a, b, c)
    return MvNormal([a,b,c], I(3))
end

likelihood = likelihoodof(splat(fwd), [0,0,0])

prior = distprod(;priors...)
posterior = PosteriorMeasure(likelihood, prior)
```

```julia
samples = bat_sample(posterior).result
```

    [ Info: Setting new default BAT context BATContext{Float64}(Random123.Philox4x{UInt64, 10}(0x0e7baf8ab291ff8c, 0x5b0485835223e643, 0x64f865a9c835b39b, 0xcbdf37fe9c162917, 0xdbf6235e76e2b0ed, 0xca60cc8ac70af658, 0x0000000000000000, 0x0000000000000000, 0x0000000000000000, 0x0000000000000000, 0), HeterogeneousComputing.CPUnit(), BAT._NoADSelected())
    ┌ Info: Using sampling algorithm TransformedMCMC{RandomWalk{TDist{Float64}}, NoMCMCProposalTuning, PriorToNormal, BAT.TriangularAffineTransform, RAMTuning, BAT.NoMCMCTempering, MCMCChainPoolInit, MCMCMultiCycleBurnin, BrooksGelmanConvergence, RepetitionWeighting{Int64}, typeof(BAT.nop_func)}
[...]

PR #514, opened 2026-06-16
Makie interactivity 

Micki-D:
[Experimental][WIP] Add an extension for Makie.jl plotting

This is based on PR https://github.com/bat/BAT.jl/pull/477 using @philippeller 's custom Makie plotting but with dedicated BAT.jl infrastructure for handling the marginalization of samples.

This implements an extension module with reactive Makie.jl recipies that offer more flexible visualization and enable live plotting of MCMC data during sampling.


BJMCox:
Rebased onto current main - am reviewing now

src/algotypes/bat_context.jl
```diff
@@ -210,14 +211,15 @@ function set_batcontext(context::BATContext)
...
     c = get_batcontext()
     s = merge(
-        (cunit = get_compute_unit(c),precision=get_precision(c), rng=get_rng(c), ad=get_adselector(c)),
-        (;kwargs...)
+        (cunit=get_compute_unit(c), precision=get_precision(c), rng=get_rng(c), ad=get_adselector(c)),
+        (; kwargs...)
     )
     adsel = _to_adsel(s.ad)
-    set_batcontext(BATContext{s.precision}(s.rng, s.cunit, adsel))
+    set_batcontext(BATContext{s.precision}(s.rng, s.cunit, adsel, c.visualizer)) # Temporary, to fix
```
BJMCox, line 222: The keyword merge accepts `visualizer`, but this call always uses `c.visualizer`. Thus `set_batcontext(visualizer=requested)` silently ignores the request. Minimal case:
```julia
old = get_batcontext()
requested = BATVisualizer(MyBackend(), nothing)
set_batcontext(visualizer=requested)
@assert get_batcontext().visualizer === requested
set_batcontext(old)
```
Could this use the merged visualizer value, like the other context fields?

src/algotypes/bat_context.jl
```diff
@@ -28,24 +28,26 @@ struct BATContext{T<:AbstractFloat,RNG<:AbstractRNG,CU<:AbstractComputeUnit,AD<:
...
     cunit::CU
     ad::AD
+    visualizer::BATVisualizer
 end
 
 export BATContext
 
 function BATContext{T}(
-    rng::RNG, cunit::CU, ad::AD
+    rng::RNG, cunit::CU, ad::AD, visualizer::BATVisualizer
```
BJMCox, line 37: Adding `visualizer` removes the documented three-argument `BATContext{T}(rng, cunit, ad)` constructor. Existing user code now gets a `MethodError`. This needs to be mentioned or have a three arg form.

src/algotypes/bat_visualizer.jl
```diff
@@ -0,0 +1,58 @@
...
+*Experimental feature, not yet part of stable public API.*
+
+Carries a visualization backend configuration and, once sampling starts, the
+live visualization state. Pass via `BATContext(visualizer = ...)`. A used
+visualizer is single-use -- create a fresh one per sampling run.
+
+Constructors:
+
+* ```BATVisualizer()```: no visualization
+* ```BATVisualizer(backend::BATVisBackend)```: e.g. `BATVisualizer(BATMakieVisualization())`
```
BJMCox, line 27: This public docstring promises `BATVisualizer(backend)`, but only the zero-argument and two-field constructors exist. `BATVisualizer(MyBackend())` gives a `MethodError`. Please either add the documented forwarding constructor or remove the documented API.

src/extdefs/makie_defs.jl
```diff
@@ -0,0 +1,359 @@
...
+
+    triagonal_config = _default_makie_triagonal_config(trace_nsteps=trace_nsteps, dark=dark)
+    diagonal_config = _default_makie_diagonal_config(dark=dark)
+
+    vis_config = BATMakieVisualization(
+        recipes,
+        vsel,
+        N_max,
+        n_batch,
+        max_buffered,
```
BJMCox, line 123: `max_buffered` may be smaller than the fixed `n_batch=50`. With `max_buffered=1`, the producer waits after one buffered row while the listener waits for 50 rows before flushing. Neither can advance. Please validate this relation or derive a reachable flush threshold.

ext/makie_impl/makie_visualizer.jl
```diff
@@ -0,0 +1,226 @@
...
+
+    # get_samples! mutates output_buffer[output_id] while flush_buffer! copies it on another task, so it must run under buffer_lock (try/finally so a throw can't leave the lock held forever).
+    lock(buffer_lock)
+    try
+        # sum(length, ...): no intermediate array on this per-step hot path.
+        n_smpls_start = sum(length, output_buffer[output_id])
+        get_samples!(output_buffer[output_id], chain_state, nonzero_weights)
+        n_smpls_end = sum(length, output_buffer[output_id])
+        n_new = n_smpls_end - n_smpls_start
+        n_buffer_samples[] += n_new
```
BJMCox, line 218: This counter measures new rows, not new MCMC steps. `checked_push!` represents a rejection by increasing the last sample weight without adding a row, so a long rejection streak never triggers a live update. Minimal invariant:
```julia
checked_push!(buffer, buffer[end])
@assert buffer.weight[end] == 2
@assert length(buffer) == old_length
```
Please count consumed steps or weight mass instead of row growth.

BJMCox (commented):
I found several correctness and lifecycle issues that should be addressed before merge. Two existing Copilot comments also still apply: the unshaped `getstring` error path uses undefined `names`, and this overhaul adds no tests. To be honest the test one is probably not super useful, but the other one deserves addressing I think.

ext/makie_impl/makie_compute_graph.jl
```diff
@@ -0,0 +1,568 @@
...
+        while l2 <= h2
+            m = (l2 + h2) >> 1
+            if Int(info[m].stepno) <= whi
+                hi = m
+                l2 = m + 1
+            else
+                h2 = m - 1
+            end
+        end
+        return lo:hi
```
BJMCox, line 75: This window assumes step numbers remain monotone, but `next_cycle!` resets them to zero while live output concatenates cycles. The binary searches can therefore drop valid rows after a cycle change. It also keeps a boundary row's full dwell weight. For `stepno=[1,101]`, `weight=[100,1]`, and window `100:101`, it plots weights `[100,1]` instead of `[1,1]`. Please preserve cycle order and clip dwell mass at both window boundaries.

ext/makie_impl/makie_kde.jl
```diff
@@ -0,0 +1,454 @@
...
+# value: density units are inverse data units, so any absolute cutoff is scale-dependent.
+const _KDE2D_DENSITY_FLOOR_FRAC = 1e-3
+
+# Weighted counterpart of KernelDensity.default_bandwidth, whose own selection
+# ignores `weights` entirely: the same robust Silverman rule with WEIGHTED std/IQR
+# (uncorrected Weights, the extension-wide convention) and n = Kish effective
+# sample size. Returns `nothing` for degenerate inputs (callers fall back to the default).
+function _weighted_kde_bandwidth(values::AbstractVector{<:Real}, weights::AbstractVector{<:Real}, alpha::Float64 = 0.9)
+    sum_w = sum(weights)
+    sum_w2 = sum(abs2, weights)
```
BJMCox, line 19: Please normalize weights before squaring them. Multiplying every weight by the same positive constant must not change a weighted KDE. Here `1e200 .* [1,2,8]` makes `sum(abs2, weights)` overflow and sends the same empirical measure down the unweighted fallback path.

ext/makie_impl/makie_stats.jl
```diff
@@ -0,0 +1,574 @@
...
+function _cov_ellipse_primitives(μ::AbstractVector, Σ::AbstractMatrix, nsigma::Real)
+    vals, vecs = eigen(Σ)
+    stds = sqrt.(clamp.(vals, 0, Inf))
+
+    ellipse_mat = vecs * nsigma * Diagonal(stds) * _COV_ELLIPSE_CIRCLE .+ μ
+    ellipse_points = Point2f.(ellipse_mat[1, :], ellipse_mat[2, :])
+
+    axes_segments = Point2f[]
+    for i in 1:2
+        direction = vecs[:, i] * stds[i] * 2 * nsigma
```
BJMCox, line 47: The ellipse uses semiaxes `nsigma * sqrt(lambda)`, while these axis segments use `2 * nsigma * sqrt(lambda)` in each direction. At `nsigma=1`, the ellipse is one standard deviation but each endpoint is two. Please use the same radius for both representations.

ext/makie_impl/makie_stats.jl
```diff
@@ -0,0 +1,574 @@
...
+    coords, w_prob = si
+    (; npoints_pdf) = config
+
+    vals = vec(coords)
+    μ = mean(vals, w_prob)
+    σ = std(vals, w_prob)
+    # A nonempty all-zero-weight vector passes the emptiness guards; std()
+    # then returns NaN, which Normal(μ, NaN) rejects with a DomainError.
+    (isfinite(μ) && isfinite(σ)) || return _empty_pdf1d_primitives()
+    dist = Normal(μ, σ)
```
BJMCox, line 539: `sigma == 0` passes the finite check. `Normal(mu, 0)` then produces an infinite density at `mu`, so the plotted values are non-finite. A constant marginal reproduces this. Please define an explicit zero-variance rendering or return empty primitives.

src/samplers/mcmc/multi_cycle_burnin.jl
```diff
@@ -49,8 +49,10 @@ function mcmc_burnin!(
 
         mcmc_states = mcmc_iterate!!(
             new_outputs, mcmc_states;
-            max_nsteps = burnin.nsteps_per_cycle,
-            nonzero_weights = nonzero_weights
+            max_nsteps=burnin.nsteps_per_cycle,
+            nonzero_weights=nonzero_weights,
+            callback=callback,
```
BJMCox, line 54: Passing the public sampling callback into `mcmc_iterate!!` adds a new per-step event, `callback(Val(:mcmc_iterate), state)`, during burn-in. A valid callback that only implements the documented cycle event now fails with `MethodError` inside the chain tasks. Please separate the internal visualizer hook from the user callback contract, or make the new event backward compatible.

src/samplers/mcmc/mcmc_sample.jl
```diff
@@ -144,11 +154,24 @@ function evalmeasure_impl(em::EvaluatedMeasure, samplingalg::TransformedMCMC, co
...
         mcmc_states;
         max_nsteps = samplingalg.nsteps,
-        nonzero_weights = samplingalg.nonzero_weights
+        nonzero_weights=samplingalg.nonzero_weights,
+        update_visualizer=true
     )
 
     @debug "Merge samples of chains and transform to original space."
 
+    if !isnothing(context.visualizer.content)
```
BJMCox, line 163: Listener shutdown only runs after successful burn-in and sampling. Any callback or sampler error jumps past this block and leaves the polling task live. A `try/finally` around the sampling lifecycle would preserve cleanup on both success and failure. A GLMakie MWE with a throwing callback confirms the task remains alive.

ext/makie_impl/makie_compute_graph.jl
```diff
@@ -0,0 +1,568 @@
...
+end
+
+# Initial per-dim axis/bin domain from prior-sample tail quantiles; not a hard
+# bound. `target` must be the ORIGINAL untransformed measure (bat_sample's own
+# `m`), NOT mcmc_target(...): mcmc states carry the PRETRANSFORMED measure,
+# giving a standard-normal-scaled domain regardless of the real prior's scale.
+function _estimate_prior_domain(target, n_dof::Integer; n_prior_samples::Integer = 2000, tail_prob::Real = 0.0015)
+    initsrc = BAT.get_initsrc_from_target(target)
+    shape = varshape(initsrc)
+    draws = [ValueShapes.unshaped(rand(initsrc), shape) for _ in 1:n_prior_samples]
```
BJMCox, line 88: These 2,000 draws use the process-global RNG. Enabling visualization therefore changes unrelated random streams and ignores the `BATContext` RNG. Please pass the context RNG through this path. Minimal check: seed the global RNG, call this function, then compare the next draw with the no-visualizer sequence.

ext/BATMakieExt.jl
```diff
@@ -0,0 +1,66 @@
...
+using BAT: BATVisualizer, BATMakieVisualization
+using BAT: _default_makie_triagonal_config, _default_makie_diagonal_config
+
+import BAT: BATVisualizer
+import BAT: init_visualizer!, register_state_for_vis!, update_visualizer_impl!
+import BAT: bat_makie_plot, bat_theme, bat_theme_dark
+
+include("./makie_impl/makie_plotting.jl")
+
+Makie.set_theme!(bat_theme())
```
BJMCox, line 54: Loading a package extension should not replace the application's global Makie theme. This changes unrelated plots merely because BAT and Makie were imported. The plotting entry points already use `with_theme`, so please keep BAT's theme scoped there.

ext/makie_impl/makie_plot_samples.jl
```diff
@@ -0,0 +1,326 @@
...
+        domain_lo=domain_lo,
+        domain_hi=domain_hi,
+        support_lo=support_lo,
+        support_hi=support_hi,
+        idxs=_clamp_vsel(vsel, n_dof, N_max),
+    )
+    return graph, n_dof
+end
+
+function Makie.convert_arguments(
```
BJMCox, line 65: Makie only forwards conversion keywords declared by `Makie.used_attributes`. Without that method, calls such as `plot(samples; vsel=[4], N_max=1)` cannot route these keywords into `convert_arguments`. Please declare the consumed attributes and add a normal `plot` entry-point test.

src/samplers/mcmc/mcmc_sample.jl
```diff
@@ -281,3 +304,89 @@ function _merge_chain_outputs(mcmc_state::MCMCState, chain_outputs::AbstractVect
...
+                for sample in outputs_B[i][j]
+                    checked_push!(merged_outputs[i][j], sample)
+                end
+            end
+        end
+    end
+
+    return merged_outputs
+end
+export _append_chain_outputs
```
BJMCox, line 338: Why does the Makie extension require this internal merge helper to become public? `using BAT` now imports `_append_chain_outputs`, despite the underscore marking it private. Please keep it unexported and import it explicitly from the extension.

src/plotting/valueshapes_utils_internal.jl
```diff
@@ -35,7 +35,7 @@ function getstring(samples::BAT.DensitySampleVector, idx::Integer)
         names = all_active_names(vs)
         return names[idx]
     else
-        throw(ArgumentError("Samples are unshaped. Key :$name cannot be matched. Use index instead."))
+        throw(ArgumentError("Samples are unshaped. Key :$names cannot be matched. Use index instead."))
```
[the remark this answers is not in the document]
BJMCox: Standalone MWE:

```julia
using BAT

samples = DensitySampleVector([[0.0]], [0.0])

observed = try
    BAT.getstring(samples, 1)
catch err
    err
end

observed isa ArgumentError || error(
    "unshaped getstring throws $(typeof(observed)) instead of ArgumentError"
)
```

src/samplers/mcmc/mcmc_sample.jl
```diff
@@ -141,21 +141,30 @@ function bat_sample_impl(m::BATMeasure, samplingalg::TransformedMCMC, context::B
...
     @info "Generate main samples using $(length(mcmc_states)) MCMC chain(s)."
     mcmc_states = mcmc_iterate!!(
         chain_outputs,
         mcmc_states;
-        max_nsteps = samplingalg.nsteps,
-        nonzero_weights = samplingalg.nonzero_weights
+        max_nsteps=samplingalg.nsteps,
+        nonzero_weights=samplingalg.nonzero_weights,
+        update_visualizer=true # TODO: MD; discuss the whole update pipeline for visualizer
     )
```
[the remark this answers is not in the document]
BJMCox: Standalone MWE:

```sh
#!/bin/sh

set -eu

base=d33ddf2bc986137c2075bb83fb4a7f4f7a816234

test_count=$(git diff --name-only "$base"...HEAD -- test | wc -l | tr -d ' ')
[ "$test_count" -gt 0 ] || {
    echo "PR 514 changes no test file" >&2
    exit 1
}
```

src/algotypes/bat_context.jl
BJMCox: Standalone MWE:

```julia
using BAT

struct MWEBackend <: BAT.BATVisBackend end

function main()
    old_context = BAT.get_batcontext()
    requested = BAT.BATVisualizer(MWEBackend(), nothing)

    try
        BAT.set_batcontext(visualizer=requested)
        observed = BAT.get_batcontext().visualizer
        observed === requested || error(
            "set_batcontext ignored visualizer: requested $(typeof(requested)), observed $(typeof(observed))"
        )
    finally
        BAT.set_batcontext(old_context)
    end
end

main()
```

src/algotypes/bat_context.jl
BJMCox: Standalone MWE:

```julia
using BAT

context = BATContext()

# This constructor remains documented in BATContext's public docstring.
BATContext{Float64}(context.rng, context.cunit, context.ad)
```

src/algotypes/bat_visualizer.jl
BJMCox: Standalone MWE:

```julia
using BAT

struct MWEBackend <: BAT.BATVisBackend end

# This constructor is documented by BATVisualizer's public docstring.
BATVisualizer(MWEBackend())
```

src/extdefs/makie_defs.jl
BJMCox: Standalone MWE:

```julia
using BAT

config = BATMakieVisualization(max_buffered=1)

flush_at = config.n_batch
block_at = ceil(Int, config.n_batch * (config.max_buffered / config.n_batch))

block_at >= flush_at || error(
    "producer blocks at $block_at buffered row, but the listener flushes at $flush_at rows"
)
```

ext/makie_impl/makie_visualizer.jl
BJMCox: Standalone MWE:

```julia
using BAT

buffer = DensitySampleVector([[0.0]], [0.0]; weight=[1])
n_rows_before = length(buffer)

BAT.checked_push!(buffer, buffer[1])
n_new = length(buffer) - n_rows_before

buffer.weight[1] == 2 || error("the MCMC dwell weight did not increase")
n_new > 0 || error("the live buffer counter sees no update when only the dwell weight changes")
```

ext/makie_impl/makie_compute_graph.jl
BJMCox: Standalone MWEs:

```julia
using BAT
using Makie

extension_module = Base.get_extension(BAT, :BATMakieExt)

struct StepInfo
    stepno::Int
end

struct Walker
    info::Vector{StepInfo}
    weight::Vector{Int}
end

Base.length(walker::Walker) = length(walker.info)

walker = Walker(StepInfo.([1, 2, 1, 2]), ones(Int, 4))
rows = collect(extension_module._step_window_rows(walker, 4, 2, 2))

rows == [2, 4] || error("the step window returned rows $rows instead of [2, 4]")
```

```julia
stepno = [1, 101]
weight = [100, 1]
wlo, whi = 100, 101

rows = findall(eachindex(weight)) do i
    stepno[i] + weight[i] - 1 >= wlo && stepno[i] <= whi
end

used_weight = weight[rows]
window_weight = [
    max(0, min(stepno[i] + weight[i] - 1, whi) - max(stepno[i], wlo) + 1)
    for i in rows
]

used_weight == window_weight || error(
    "the two-step window uses weights $used_weight instead of $window_weight"
)
```

ext/makie_impl/makie_kde.jl
BJMCox: Standalone MWE:

```julia
using Statistics
using StatsBase

function bandwidth(values, weights)
    sum_w = sum(weights)
    sum_w2 = sum(abs2, weights)
    (isfinite(sum_w) && isfinite(sum_w2) && sum_w > 0 && sum_w2 > 0) || return nothing
    n_eff = sum_w^2 / sum_w2
    n_eff <= 1 && return nothing

    w = Weights(weights)
    q25, q75 = quantile(values, w, [0.25, 0.75])
    width = min(std(values, w), (q75 - q25) / 1.34)
    return 0.9 * width * n_eff^(-0.2)
end

values = [0.0, 1.0, 100.0]
weights = [1.0, 2.0, 8.0]

bandwidth(values, weights) == bandwidth(values, 1e200 .* weights) || error(
    "a common finite weight scale changes the bandwidth path"
)
```

ext/makie_impl/makie_stats.jl
BJMCox: Standalone MWE:

```julia
eigenvalues = [1.0, 4.0]
nsigma = 1.0

ellipse_radii = nsigma .* sqrt.(eigenvalues)
axis_endpoint_radii = 2 * nsigma .* sqrt.(eigenvalues)

axis_endpoint_radii == ellipse_radii || error(
    "axis endpoints use radii $axis_endpoint_radii for ellipse radii $ellipse_radii"
)
```

ext/makie_impl/makie_stats.jl
BJMCox: Standalone MWE:

```julia
using Distributions

μ = 3.0
σ = 0.0
x = collect(LinRange(μ - 1, μ + 1, 3))
y = pdf.(Normal(μ, σ), x)

all(isfinite, y) || error("the zero-variance Gaussian overlay contains $y")
```

src/samplers/mcmc/multi_cycle_burnin.jl
BJMCox: Standalone MWE:

```julia
using BAT
using Distributions

cycle_callback(::Val{:mcmc_burnin}, ::AbstractVector) = nothing

bat_sample(
    Normal(),
    TransformedMCMC(
        nchains=2,
        nsteps=10,
        store_burnin=true,
        callback=cycle_callback,
    ),
)
```

src/samplers/mcmc/mcmc_sample.jl
BJMCox: Standalone MWE:

```julia
using BAT
using Distributions
using GLMakie

visualizer = BATVisualizer(BATMakieVisualization(max_buffered=10^6))
algorithm = TransformedMCMC(
    nchains=2,
    nsteps=10,
    store_burnin=true,
    callback=(args...) -> error("stop sampling"),
)

try
    bat_sample(Normal(), algorithm, BATContext(visualizer=visualizer))
catch
end

sleep(0.2)
leaked = visualizer.content.is_live[] && !istaskdone(visualizer.content.listener_task[])

visualizer.content.is_live[] = false
wait(visualizer.content.listener_task[])

leaked && error("a sampling error left the Makie listener alive")
```

ext/makie_impl/makie_compute_graph.jl
BJMCox: Standalone MWE:

```julia
using BAT
using Distributions
using Makie
using Random

extension_module = Base.get_extension(BAT, :BATMakieExt)

Random.seed!(1)
expected = rand()

Random.seed!(1)
extension_module._estimate_prior_domain(
    BAT.BATDistMeasure(Normal()),
    1;
    n_prior_samples=2,
)
observed = rand()

observed == expected || error("domain estimation consumed process-global RNG state")
```

ext/BATMakieExt.jl
BJMCox: Standalone MWE:

```julia
using Makie

Makie.set_theme!(fontsize=11)
using BAT

Makie.current_default_theme()[:fontsize][] == 11 || error(
    "loading BAT's Makie extension replaced the process-global Makie theme"
)
```

ext/makie_impl/makie_plot_samples.jl
BJMCox: Standalone MWE:

```julia
using BAT
using Distributions
using LinearAlgebra
using Makie

samples = bat_sample(
    MvNormal(zeros(4), I),
    IIDSampling(nsamples=20),
).result

plot(samples; vsel=[4], N_max=1)
```

src/samplers/mcmc/mcmc_sample.jl
BJMCox: Standalone MWE:

```julia
module ImportProbe
using BAT

isdefined(@__MODULE__, :_append_chain_outputs) && error(
    "using BAT imports the underscore-private helper _append_chain_outputs"
)
end
```

PR #526, opened 2026-08-15
Fix logarithmic-weight resampling

BJMCox:
Avoids underflow in `RandResampling` with `ULogarithmic` weights by sampling in log space. Preserves ordinary-weight behavior, zero-sample requests, and immutable weight storage. Adds regression tests.


oschulz:
Thanks!

Issue #528, opened 2026-08-15
Modernize MCMC convergence diagnostics


BJMCox:
`BrooksGelmanConvergence` uses the older variance-based Brooks-Gelman statistic and is the default for `TransformedMCMC`.

Modern practice recommends rank-normalized split R-hat, folded R-hat, and bulk/tail effective sample sizes. These better detect non-stationarity and tail problems.

Consider adding these diagnostics and then reassessing the default.

Reference: Vehtari et al. (2021), https://doi.org/10.1214/20-BA1221

PR #530, opened 2026-08-15
Add rank-normalized R-hat convergence checker

BJMCox:
Addresses the R-hat part of #528.

Add opt-in `RankNormalizedRhatConvergence` with a default threshold of 1.01. Compute rank-normalized and folded split R-hat across independent chains, preserving walker identity, draw order, and integer repetition weights without expanding draws. Return a nonconverged result for short, constant, or nonfinite trajectories.

## Convergence types

Document the public MCMC convergence hierarchy:

```text
ConvergenceTest
├── AssumeConvergence
└── MCMCConvergenceTest
    ├── SingleChainMCMCConvergenceTest
    └── MultiChainMCMCConvergenceTest
        ├── GelmanRubinConvergence
        ├── BrooksGelmanConvergence
        └── RankNormalizedRhatConvergence
```

All three existing diagnostics share the multi-chain sample adapter and one-chain sampler guard. The single-chain abstract type has no concrete diagnostics yet. `BrooksGelmanConvergence` remains the default.

### DIFF
--- a/docs/src/stable_api.md
+++ b/docs/src/stable_api.md
@@ -126,6 +126,7 @@ NormalBased
 UniformBased
 RAMTuning
 RandomWalk
+RankNormalizedRhatConvergence
 RandResampling
 RepetitionWeighting
 RiceBinning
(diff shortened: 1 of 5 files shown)


oschulz:
May be spun out into another package.

PR #540, opened 2026-08-16
Add configurable credible intervals to sample reports

### LINKED ISSUE #381: Custom CIs and/or printing all CI values via bat_report
Hello! 

I am a new user of BAT, and currently use bat_report(samples) for the credible intervals to be printed on screen. However, that only prints out the 68% CI, not the others.

Also, is it possible to change the default CIs ? 

If you could add some functions to customise CI quantiles and print out all CIs on screen (all CIs shown in the marginalised plot), it could be of great help.

BJMCox:
Fixes #381.

Add configurable probability masses and interval modes to sample reports through `smallest_credible_intervals(...; p, mode)`. Reports show all disjoint components, support `mode = :connected` and full mass, and retain the default masses and deprecated `bat_report` keyword forwarding.

Main code: +42/-7.

### DIFF
--- a/docs/src/tutorial_lit.jl
+++ b/docs/src/tutorial_lit.jl
@@ -261,6 +261,10 @@ println("Covariance: $par_cov")
 using LazyReports
 lazyreport(samples)
 
+# Specify probability masses for the smallest intervals shown in the report;
+# these use the same histogram selection and edges as the marginal plots:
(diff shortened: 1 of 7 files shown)


BJMCox:
Surprisingly tricky but this seems to work?

oschulz:
Let's think about what kind of API we need for this in LazyReports so that such user preferences can propagate.

PR #541, opened 2026-08-16
Fix weighted bridge-sampling error estimates

### LINKED ISSUE #531: Fix negative bridge-sampling error estimates for fractional weights
`BridgeSampling` still throws for fractional weights after #527. The proposal-count conversion succeeds, but the error estimate can become negative before `sqrt`.

```julia
using BAT, Distributions, Random

ctx = BATContext(rng = Xoshiro(165))
dist = MvNormal(zeros(1), ones(1))
samples = bat_sample(dist, IIDSampling(nsamples = 3), ctx).result
samples = DensitySampleVector(samples.v, samples.logd; weight = fill(0.75, 3))

bat_integrate(EvaluatedMeasure(dist; samples), BridgeSampling(), ctx)
```

On `d33ddf2b`:

```text
ERROR: DomainError with -0.22960883840259716:
sqrt was called with a negative real argument
```

`EffSampleSizeFromAC` returns `-2.0` for the retained samples. The bridge error calculation then evaluates `0.884925131820357 - 1.1145339702229544` before taking the square root. Fractional-weight rounding and the weighted covariance make the result weight-scale dependent. Changing only the weights to `1.0` returns a finite result.

Please implement a valid weighted relative-MSE estimate with finite positive ESS, without merely clamping negative results. Add deterministic integer, fractional, normalized, unequal, and common-scale-equivalent weight tests.

BJMCox:
Fixes #531.

Normalize bridge-sampling weights and use weight effective sample sizes for proposal sizing, weighted moments, and the relative-MSE estimate. Keep invalid autocorrelation fallback local to bridge sampling. Weighted resampling is invariant under common weight scaling.

Main code: +53/-23.


BJMCox:
Superseded by #564, now merged into main.

PR #543, opened 2026-08-16
Preserve concrete types in MCMC proposal mixtures

### LINKED ISSUE #533: Make MCMCMultiProposal defaults constructible
`MCMCMultiProposal` only accepts proposal vectors, but its default proposal value is a tuple.

```julia
using BAT

MCMCMultiProposal((MetropolisHastings(),), [1])
```

On `d33ddf2b`:

```text
ERROR: MethodError: no method matching MCMCMultiProposal(
    ::Tuple{RandomWalk{Float64, Tuple{Float64, Float64}, TDist{Float64}}},
    ::Vector{Int64},
)

Closest candidates are:
  MCMCMultiProposal(!Matched::P, ::R) where
    {P<:(Vector{<:BAT.MCMCProposal}), ...}
```

The equivalent vector input constructs successfully. The type parameter requires `P<:Vector`, while the zero-argument default is `(RandomWalk(), HamiltonianMC())`. That default also fails earlier when the optional AdvancedHMC backend is not loaded.

Please choose one public input representation, normalize accepted inputs, and make the zero-argument default usable without an unloaded optional backend. Validate non-empty proposals and picking-rule lengths. Add default, tuple/vector, and `Categorical` constructor tests.

BJMCox:
Fixes #533.
Main changes (this is operating on the multi proposal stuff)

- Vectors -> tuples: Julia now has each proposal’s and tuner’s exact type.
- Repeated loops -> shared helpers: less duplicate code.
- Dynamic indexing -> specialised branches: each selected proposal runs with known types.
- Tuner fields and inlining: retain concrete adaptor types and remove extra MALA boxing.
- Default -> RandomWalk only: avoids requiring automatic differentiation by default.

Source: +101/-114. Tests: +26/-0, comprising the default regression and one mixed sampling case.

### DIFF
--- a/src/samplers/mcmc/mcmc_multiproposal.jl
+++ b/src/samplers/mcmc/mcmc_multiproposal.jl
@@ -13,13 +13,27 @@ Fields:
 
 $(TYPEDFIELDS)
 """
-@with_kw struct MCMCMultiProposal{
+struct MCMCMultiProposal{
     P<:Vector{<:MCMCProposal},
     R<:Union{Vector{<:Integer}, Categorical}
 }<:MCMCProposal
-    # TODO: MD, should we put a default tuple of proposals, if so, what should it be?
-    proposals::P = MCMCProposal[RandomWalk(), HamiltonianMC()]
-    picking_rule::R = Categorical(1/length(proposals) .* ones(length(proposals)))
+    proposals::P
+    picking_rule::R
+end
+
+function MCMCMultiProposal(
+    proposals::Tuple{Vararg{<:MCMCProposal}},
+    picking_rule::Union{Vector{<:Integer}, Categorical}
+)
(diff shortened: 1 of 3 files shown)


oschulz:
We want type stability and proposal types will typically be mixed.

BJMCox:
Proposal states also lose component types in [`mcmc_multiproposal.jl`](https://github.com/bat/BAT.jl/blob/4311d51a536e396b378193eea1642994fe7553aa/src/samplers/mcmc/mcmc_multiproposal.jl#L78), and tunings/tuner states do so in [`mcmc_multiproposal_tuner.jl`](https://github.com/bat/BAT.jl/blob/4311d51a536e396b378193eea1642994fe7553aa/src/samplers/mcmc/mcmc_tuning/mcmc_multiproposal_tuner.jl#L16).
I am going to take a look at these as well.

PR #564, opened 2026-08-23
Major upgrade

oschulz:
### Breaking changes

* The new function `evalmeasure(target, algorithm, context)` is now BAT's
  central evaluation mechanism. It returns an `EvaluatedMeasure` that
  annotates the target with the knowledge the algorithm produced (samples,
  approximations, modes, DOF, mass) and can be passed to further algorithms,
  which accumulate knowledge on it.

* `bat_sample`, `bat_findmode` and `bat_integrate` are now thin wrappers
  around `evalmeasure` and return only `(result = ..., ...)`. They no longer
  return an `evaluated` field or splat backend-specific extras into their
  return value. Use `evalmeasure` directly to obtain an `EvaluatedMeasure`;
  backend-specific information is available via `evalinfo`.

* `EvaluatedMeasure` has been redesigned: it carries a transformed-space
  view (`transform_intent`, `f_transform`) so that follow-up evaluations can
  reuse sample transports and measure representations. Its fields are
  public for reading; the keyword constructor is the update API. The new
  accessor functions `empiricalof`, `samplesof`, `approxof`, `samplegenof`,
  `getess`, `evalinfo` and `unevaluated` are the stable read API.

* The new public type `DensitySampleMeasure` represents an empirical
  measure backed by a `DensitySampleVector` (convertible in both
  directions), with optional DOF, effective sample size and mass. Sampling
  results wrap their samples in it. It has mass one by default and always
  reflects the live sample weights.

* Statistics of weighted samples (`var`, `std`, `cov` of
  `DensitySampleVector`) now use uncorrected probability-weight moments,
  matching the empirical-measure semantics. Values change slightly for
  weighted samples.

* Effective-sample-size estimation is provenance-driven:
  `EffSampleSizeFromAC` reconstructs the exact per-walker ordered sequences
  from MCMC sample ids (even after merging and resampling) and pools
  independent chains and walkers by their weight-mass fractions;
  `KishESS` is the default for nonuniformly weighted samples without such
  provenance.

* Measure masses and integrals are stored on a canonical logarithmic scale
  (`LogarithmicNumbers.ULogarithmic`) uniformly across all integration and
  sampling backends.

* `HamiltonianMC` is now a native NUTS implementation. The AdvancedHMC
  backend and package extension have been removed, as has `StanLikeTuning`.

* `MALA` has been reworked, with an exact Hastings correction and
  dual-averaging step-size adaptation.

* Gradient-based MCMC proposals (`HamiltonianMC`, `MALA`) now default to
  the new `FisherTransformTuning`; `RandomWalk` keeps `RAMTuning`.

* Mode estimation has been disentangled from the optimization backends:
  the new `TransformedMaxDensity` combines a pure optimization backend
  (`OptimAlg`, `OptimizationAlg`) with a space transformation and
  initializer, and serves `bat_findmode` and `bat_bgml`. Raw optimizers
  work directly, e.g. `bat_findmode(target, Optim.NelderMead())`.
  `MaxDensityNelderMead`, `MaxDensityLBFGS`, `NelderMeadOpt` and `LBFGSOpt`
  have been deprecated.

* Renamed (with deprecated aliases): `AbstractTransformTarget` is now
  `TransformIntent`, `PriorToNormal` is `NormalBased`, `PriorToUniform` is
  `UniformBased`, and `OrderedResampling` is `SystematicResampling`.

* `IIDSampling` now throws an error for targets that cannot guarantee
  independent draws (e.g. sample-backed empirical measures) - use a
  resampling algorithm instead.

* `set_batcontext` now sets the default context process-wide instead of
  per-task (see below).

* No longer exported: `AbstractSampleGenerator`, `AnySampleable`,
  `bat_ensemble_initvals` and several internal MCMC state and tuner types.

* BAT now requires Julia >= 1.10 and depends on MatrixShapedOperators and
  ScopedSettings.

### New features

* Knowledge chaining via `evalmeasure`: evaluated measures can be passed to
  further algorithms, which reuse existing samples, transformations and
  approximations where possible.

* `FisherTransformTuning`: affine space-transform tuning by matching the
  score of the target (Fisher divergence), with low-rank structure
  detection and a drift-based commit schedule (`DriftCommitSchedule`),
  including regimes where the dense tuner would be too expensive.

* `SystematicResampling` now implements true systematic resampling
  (single stratified uniform, order-preserving, typically lower variance
  than multinomial resampling) and preserves MCMC sample-id provenance;
  `RandResampling` clears it.

* The default computational context is managed by a
  `ScopedSettings.ScopedSetting`, exported as `default_batcontext`:
  `set_batcontext` installs a process-wide default visible to all tasks,
  `ScopedSettings.with` binds a context for a dynamic scope (inherited by
  tasks started within it), and without an override each access yields a
  fresh context seeded from `Random.default_rng()`, so BAT results are
  reproducible via `Random.seed!`.

* Experimental: Pathfinder-based initialization of adaptive space
  transformations (`BAT.PathfinderTransformInit`) and low-rank affine
  transforms (`BAT.LowRankAffineTransform`).

* `batalgorithm` (experimental) normalizes user-provided algorithm-like
  objects into BAT algorithms.

* Many robustness fixes throughout: sample-weight arithmetic, resampling,
  ESS estimation and bridge sampling are now safe at extreme weight scales
  and invariant under global weight rescaling.

Assisted by generative AI.


BJMCox:
[`d5370822`](https://github.com/bat/BAT.jl/commit/d5370822fbe5bab0dfc7d3724e6c44f7e14693eb) changed quantiles from `FrequencyWeights` to `ProbabilityWeights`, so compressing repeats changed the result for the same empirical measure.

[`ebf96b03`](https://github.com/bat/BAT.jl/commit/ebf96b03) uses the inverse weighted ECDF and preserves scalar types. Weights `[1,2]` and `[2,4]` both normalize to masses `[1/3,2/3]`, so they describe the same empirical distribution. Their Julia type-7 expansions have 25% quantiles `0.5` and `0.25`. Matching both would let an arbitrary global weight scale change the reported quantile without changing the distribution.

BJMCox:
I have opened #567 to handle this separately.

The current low-rank Fisher update can mistake noise for correlation on non-Gaussian targets, then keep changing the geometry until tuning fails. The Gaussian tests miss this because the relevant terms cancel out.

The fix keeps low-rank learning, but limits it to one attempt and checks the result on held-out samples. The correction is only retained when its off-diagonal part improves the fit. For HMC, it is applied after that check. For MALA, it is applied while the step size is tuned, then reverted if the check fails.

Invalid fits now fall back safely, and short burn-in cycles can try again later. I have added null, correlated-target, lifecycle, and end-to-end tests.

BJMCox:
The transformed empirical cache could return stale weights after live reweighting. `DensitySampleMeasure` keeps its weights live, but transforming the samples copied the weight vector. Changing the main weights from `[1, 1]` to `[0, 2]` therefore left the cached transformed side at `[1, 1]`.

The two representations then described different empirical measures, so transformed-space code could use the wrong atom masses.

Fixed by making BAT-produced empirical pairs share one weight vector. Supplied pairs still keep their object identity; when their weights are separate, we skip the cache and rebuild from the current main side. Standalone sample transforms still copy their weights.

BJMCox:
Systematic resampling can repeat the same MCMC state. We treated those repeated IDs as lost provenance, so ESS used storage order instead of reconstructing each walker.

This keeps repeated IDs when their values match, while rejecting conflicting duplicates. ESS now follows walker history after resampling.

BJMCox:
Okay this is everything that I think *needs* to be done - I have other things that can be done as PRs after this rolls into a major release

src/statistics/effective_sample_size.jl
```diff
@@ -180,13 +180,22 @@ function bat_eff_sample_size_impl(smpls::DensitySampleVector, algorithm::EffSamp
...
+# repeats remain valid after order-preserving systematic resampling;
+# conflicting values for one process id do not:
 function _has_process_provenance(unshaped_smpls::DensitySampleVector)
     info = unshaped_smpls.info
     eltype(info) <: MCMCSampleID || return false
-    return allunique((id.chainid, id.walkerid, id.chaincycle, id.stepno) for id in info)
+    allunique((id.chainid, id.walkerid, id.chaincycle, id.stepno) for id in info) && return true
+
+    seen = Dict{Tuple{Int32,Int32,Int32,Int64},Int}()
+    for i in eachindex(info)
```
BJMCox, line 271: Can this be done using a broadcast? We want to be GPU friendly. Apply this comment generally - all for loops should be broadcasts where possible

NEWS.md
```diff
@@ -1,6 +1,48 @@
 BAT.jl Release Notes
 ====================
 
+BAT.jl v5.0.0
+-------------
+
+### Migration from v4
```
BJMCox: Keep to format of the older ones:
list breaking changes first
note that there is new eval measure but old bat sample still works

src/statistics/effective_sample_size.jl
```diff
@@ -137,11 +146,88 @@ end
 export EffSampleSizeFromAC
 
 
+_ac_ess_singleton(::AbstractVector{Float16}) = one(Float32)
```
BJMCox: Weird promotion - check if this is required and occurs everywhere

src/samplers/bat_sample.jl
```diff
@@ -40,6 +40,9 @@ export IIDSampling
 
 function evalmeasure_impl(em::EvaluatedMeasure, algorithm::IIDSampling, context::BATContext)
     m = unevaluated(em)
+    m isa DensitySampleMeasure && throw(ArgumentError(
```
BJMCox: What is this - check on it and either prune or make a specialisation. We do not want isa stuff

PR #566, opened 2026-08-26
Support ArraysOfArrays v1

oschulz:
Adds compatibility with ArraysOfArrays v1 while keeping support for v0.5 and v0.6.

`nestedview` and the nested-array constructors are deprecated in ArraysOfArrays v1, the `VectorOfSimilarVectors` constructor and `convert` used instead work identically on all supported versions.

Running CI against ArraysOfArrays v1 needs the compatible releases of EmpiricalDistributions (oschulz/EmpiricalDistributions.jl#35), ValueShapes (oschulz/ValueShapes.jl#91) and BATTestCases (bat/BATTestCases.jl#20), until then the resolver picks ArraysOfArrays v0.6.

Includes a patch version bump.

Created by generative AI.


oschulz:
Closes #568.

PR #567, opened 2026-08-26
Validate low-rank Fisher updates

BJMCox:
Follow-up to #564.

The rolling low-rank Fisher update can mistake noise for correlation on non-Gaussian targets and keep changing the geometry until tuning fails. This keeps low-rank learning, but limits it to one attempt and checks the result on held-out samples.

The correction is retained only when its off-diagonal part improves the fit. For HMC, it is applied after that check. For MALA, it is applied while the step size is tuned, then reverted if the check fails.

Invalid fits fall back safely, and short burn-in cycles can try again later. The change adds null, correlated-target, lifecycle, and end-to-end tests.

The full headless suite passes 21,698 tests.


oschulz:
Please merge at your convenience.

PR #570, opened 2026-08-30
Add SliceSampling.jl adapter

BJMCox:
Adds optional SliceSampling.jl support for BAT targets, with stepping-out updates by default and configurable upstream samplers. The adapter uses BAT transforms, initialization, and context RNG, and preserves sample densities and transition diagnostics.

Initial values must have finite target log-density. This prevents a zero-density starting point from entering retained output when burn-in is zero.

Local validation on Julia 1.12.7: the full BAT suite passes 1,155/1,155. Five optional checks cover a conjugate posterior, shaped values with a custom sampler, density bookkeeping, burn-in order and seeded replay, and invalid initialization.


src/extdefs/slicesampling_defs.jl
```diff
@@ -0,0 +1,41 @@
...
+
+"""
+    struct SliceMCMCSampling <: AbstractSamplingAlgorithm
+
+*Experimental feature, not part of stable public API.*
+
+Sample a transformed target with SliceSampling.jl. The default sampler applies
+stepping-out slice updates in random coordinate order.
+
+The initial value must have finite target log-density. For restricted likelihood
```
oschulz: > The initial value must have finite target log-density.  ...

Too much detail?

oschulz (changes requested):

src/extdefs/slicesampling_defs.jl
BJMCox: Indeed - trimmed

PR #571, opened 2026-08-30
Add EllipticalSliceSampling.jl adapter

BJMCox:
Adds optional EllipticalSliceSampling.jl support for BAT posterior measures. BAT transforms the prior to a Gaussian, passes the transformed likelihood separately, and returns samples with full target log-densities and autocorrelation effective sample size.

Initial values must have finite target log-density. This prevents the upstream sampler from accepting impossible proposals after a zero-density start. Nested posterior priors are unsupported.

Local validation on Julia 1.12.7: the full BAT suite passes 1,155/1,155. Four optional checks cover a conjugate posterior, burn-in order and seeded replay, densities under a shaped non-Gaussian prior, and invalid initialization.


src/extdefs/ellipticalslicesampling_defs.jl
```diff
@@ -7,6 +7,11 @@
 
 Sample a posterior with EllipticalSliceSampling.jl. BAT transforms the prior to a
 standard Gaussian and evaluates the likelihood separately, as required by the method.
+The prior must support direct transformation to a Gaussian distribution. Nested
+posterior priors are not supported.
+
+The initial value must have finite target log-density. For restricted likelihood
```
oschulz: > The initial value must have finite target log-density. 

True, but do we need that in docstrings? We'd have to put in into a lot of places, right=

oschulz (changes requested):

src/extdefs/ellipticalslicesampling_defs.jl
BJMCox: Indeed we would. Trimmed as unnecessary.

oschulz (approved):

oschulz (approved):

